PR #1, opened 2023-11-16
Add uptime, mem usage and cpu usage

DurgNomis-drol:


tplinkrouterc6u/client.py
```diff
@@ -132,10 +132,17 @@ async def _get_firmware(self) -> Firmware:
...
     async def _get_status(self) -> Status:
+
+        def _calc_cpu_usage(data: dict) -> float:
+            return (data['cpu_usage'] + data['cpu1_usage'] + data['cpu2_usage'] + data['cpu3_usage']) / 4
+
         data = await self._get_data('admin/status?form=all')
         status = Status
         status.devices = []
         status.macaddr = data['lan_macaddr']
+        status.wan_ipv4_uptime = data['wan_ipv4_uptime']
```
AlexandrErohin, line 143: I dont see parameter `wan_ipv4_uptime` in my router response. Im afraid not all routers it pass

tplinkrouterc6u/dataclass.py
```diff
@@ -30,4 +30,10 @@ class Status:
     guest_5g_enable: bool
     wifi_2g_enable: bool
     wifi_5g_enable: bool
+    wan_ipv4_uptime: int
+    mem_usage: float
+    cpu_usage: float
+    cpu1_usage: float
```
AlexandrErohin, line 36: `cpu1_usage`, `cpu2_usage` and `cpu3_usage` are not needed as you calculate it in `cpu_usage`

tplinkrouterc6u/dataclass.py
```diff
@@ -30,4 +30,10 @@ class Status:
     guest_5g_enable: bool
     wifi_2g_enable: bool
     wifi_5g_enable: bool
+    wan_ipv4_uptime: int
+    mem_usage: float
```
AlexandrErohin, line 34: could you add new fields `mem_usage` and `cpu_usage` to [ README.md ](https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/README.md?plain=1#L76)

tplinkrouterc6u/client.py
AlexandrErohin, line 143: I dont see parameter `wan_ipv4_uptime` in my router response. Im afraid not all routers it pass

tplinkrouterc6u/dataclass.py
AlexandrErohin, line 36: `cpu1_usage`, `cpu2_usage` and `cpu3_usage` are not needed as you calculate it in `cpu_usage`

AlexandrErohin (commented):
@DurgNomis-drol Thank you!
Could you check my comments please?

tplinkrouterc6u/client.py
DurgNomis-drol: What if I make it return `None` if it isn't avaliable? 

Mine is an Archer AX10

AlexandrErohin:
@DurgNomis-drol Thanks!

DurgNomis-drol:
We should use `.get()` instead, as this will return `None` if it is not in the data, instead of throwing an error

PR #3, opened 2023-11-20
Use get instead

DurgNomis-drol:


AlexandrErohin:
I have just checked all fields and realize, that it requires to change also dataclases and the way they are used further.
So I think changing for all fields supporting None is not good. These are fields what is important and if they cannot be found in response, thats the whole integration in not supported, and should be reviewed.

I have found the error with previous changes - we forgot to change [in ](https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/1/files#diff-104a414ee8d3053b27f57268112541721647509658676c8facb222d3fabd73c1R33) that wan_ipv4_uptime is int | None

AlexandrErohin:
@DurgNomis-drol could you tell, what is Hardware Version of your router? and Firmware Version please
You have Archer AX10 right ?

DurgNomis-drol:
```
Archer AX10
Firmware: 1.3.9 Build 20230728 Rel. 45716(4555)
Hardware: Archer AX10 v1.0
```

I haven't forgotten about this PR, just very busy at the moment

AlexandrErohin:
@DurgNomis-drol Thanks for the info!
Strange, I thought that Archer AX10 v1.0 API and Archer AX55 v1.0 API are the same. But here https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/5 says that integration with Archer AX55 v1.0 doesnt work

Does your router UI looks like this ? https://emulator.tp-link.com/Archer_AX10v1_US_simulator/index.html

DurgNomis-drol:
Yes it does

AlexandrErohin:
@DurgNomis-drol could you give the example of the wan_ipv4_uptime parameter? I would like to add the sensor for it, but I dont know how it looks like. Thanks

AlexandrErohin:
The client was rebuilded

PR #8, opened 2024-01-26
configurable timeout

gordonaspin:
code as-is would throw timeout exception with my Archer A6 when calling get_status(). I added timeout to TpLinkRouter defaulted to 60 seconds


AlexandrErohin:
@gordonaspin Thank you! 60 seconds too much I think. Could you test on your router timeout for 10 seconds?

gordonaspin:
10s works. It takes what it takes ... !

AlexandrErohin:
@gordonaspin Great. Could you also update README.md with you new additions? Also, what is version of your router? Archer A6 v2?

gordonaspin:
Mine is:

firmware version: 1.2.1 Build 20220714 rel.75573(5553)
hardware version: Archer A7 v5.0

I'll update readme with my second commit. I added APIs to get dhcp lease info and a couple of others. I also used macaddress and ipaddress libs to have a client program be able sort by ip (vs a lexical sort). I'll send another pull request maybe tomorrow.

gordonaspin:
I updated the README and cleaned up naming conventions, added requirements.txt.

requirements.txt
AlexandrErohin, line 1: does it needed if there is install_requires in setup.py?

tplinkrouterc6u/client.py
```diff
@@ -151,22 +167,64 @@ def _calc_cpu_usage(data: dict) -> float | None:
...
+        ipv4_status.remote = self._str2bool(data['remote'])
+
+        return ipv4_status
+
+    def _get_ipv4_reservations(self) -> [IPv4Reservation]:
+        ipv4_reservations = []
+        data = self._get_data('admin/dhcps?form=reservation', 'operation=load')
+
+        for item in data:
+            ipv4_reservations.append(IPv4Reservation(macaddress.EUI48(item['mac']), ipaddress.IPv4Address(item['ip']), item['comment'], self._str2bool(item['enable'])))
```
AlexandrErohin, line 201: Is there a way to implement sort method in IPv4Reservation to sort by ip without using macaddress and ipaddress? I made this client for Home Assistant component https://github.com/AlexandrErohin/home-assistant-tplink-router and Home Assistant work with ip and macaddress as string, so with this additions it would need to cast every time to str

gordonaspin: I kept the original interface properties to the dataclasses and added @property decorators to return the macaddress/ipaddress datatypes. I also made all the xxxStatus dataclasses with __init__() to return objects rather than use class attributes (this keeps everything following the same pattern). I also added a test.py utility to validate my changes fit your HA component use. The utility also runs through a list of queries in queries.txt to call every cgi-bin form and log the dictionary result to a corresponding file in the logs folder. I captured every click in the UI using the chrome developer tools and saved to a .har file. Then used that to find all the "/admin/foo?form=bar" URIs and call them using the client query(query, operation) API. Hope you like the additions.

AlexandrErohin:
Thank you!

AlexandrErohin:
As 10 sec works for you - Ill reduce 60 sec to 10 ok?

PR #24, opened 2024-05-17
Tp-link C1200 V2 error in request to /admin/smart_network?form=game_accelerator&operation=loadDevice 

swwgames:
tp link c1200 v2 does not have ```/admin/smart_network?form=game_accelerator&operation=loadDevice```

The router responds with: ```No root node was registered, this usually happens if no module was installed. Install luci-mod-admin-full and retry. If the module is already installed, try removing the /tmp/luci-indexcache file.```

This causes an error, because json is expected.
I added self._smart_network = False to class TplinkC1200Router


AlexandrErohin:
@swwgames Thank you!

PR #26, opened 2024-05-19
Add down_speed and up_speed to Device and add data from Deco.

nicholas-l:
The data was already in the Deco tests so this should be an easy addition. In case we don't have the data in the returned JSON, we are using `.get` which will return None if it doesn't exist so this change should not break anything.

This is in reference to https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/60


AlexandrErohin:
@nicholas-l Thanks.
What is your hardware version?

nicholas-l:
Thanks for merging and the changes to Home Assistant package! It is a Deco M5 V3

PR #54, opened 2024-11-11
Adds support for EX511 router

krlosrenan:
This pull request adds support for the EX511 router, I tested it with the Home Assistant integration and it seems to be working.

Here I'm using the following version:
Firmware Version:0.7.0 3.0.0 v607e.0 Build 240930 Rel.11206n
Hardware Version:EX511 v2.0 00000000

If you want me to change anything, please let me know.

### DIFF
--- a/tplinkrouterc6u/client.py
+++ b/tplinkrouterc6u/client.py
@@ -1497,6 +1497,16 @@ class TPLinkEXClient(AbstractRouter):
 
     def supports(self) -> bool:
         try:
+            supported_routers = ['EX511', 'EX220']
+
+            url = self._get_url('js/oid_str.js')
+            (code, response) = self._request(url)
+            assert code == 200
+            
+            pattern = r'\s*var\s+INCLUDE_BOOT_AGINETCONFIG_MODEL\s*=\s*"([^"]*)"'
+            match = search(pattern, response)
+            assert match and match.group(1) in supported_routers
+
             self._req_rsa_key()
             return True
         except AssertionError:
@@ -1555,8 +1565,8 @@ class TPLinkEXClient(AbstractRouter):
             self.ActItem(self.ActItem.GL, 'DEV2_ADT_LAN', attrs=['MACAddress', 'IPAddress']),
             self.ActItem(self.ActItem.GL, 'DEV2_ADT_WAN',
                          attrs=['enable', 'MACAddr', 'connIPv4Address', 'connIPv4Gateway']),
-            self.ActItem(self.ActItem.GL, 'DEV2_ADT_WIFI_COMMON', attrs=[]),
-            self.ActItem(self.ActItem.GL, 'DEV2_HOST_ENTRY', attrs=[]),
+            self.ActItem(self.ActItem.GL, 'DEV2_ADT_WIFI_COMMON', attrs=['primaryEnable', 'guestEnable']),
+            self.ActItem(self.ActItem.GL, 'DEV2_HOST_ENTRY', attrs=['active', 'X_TP_LanConnType', 'physAddress', 'IPAddress', 'hostName']),
             self.ActItem(self.ActItem.GO, 'DEV2_MEM_STATUS', attrs=['total', 'free']),
             self.ActItem(self.ActItem.GO, 'DEV2_PROC_STATUS', attrs=['CPUUsage']),
         ]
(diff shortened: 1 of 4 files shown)


AlexandrErohin:
Hi @krlosrenan 
Thank you very much for this!
Could you create file test_client_ex.py and write tests for methods please ? like for Tp-Link MR here - https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/test/test_client_mr.py

Tests are needed for supporting the router

tplinkrouterc6u/client.py
```diff
@@ -1428,11 +1429,509 @@ def _prepare_data(self, data: str, is_login: bool) -> tuple[str, str]:
...
+        # format expected raw request data
+        return signature, encrypted_data
+
+
 class TplinkRouterProvider:
     @staticmethod
     def get_client(host: str, password: str, username: str = 'admin', logger: Logger = None,
                    verify_ssl: bool = True, timeout: int = 30) -> AbstractRouter:
-        for client in [TplinkC5400XRouter, TPLinkMRClient, TplinkC6V4Router, TPLinkDecoClient, TplinkRouter]:
+        for client in [TPLinkEXClient, TplinkC5400XRouter, TPLinkMRClient, TplinkC6V4Router, TPLinkDecoClient, TplinkRouter]:
```
AlexandrErohin, line 1934: Could you change it to 
```
for client in [TplinkC5400XRouter, TPLinkMRClient, TPLinkEXClient, TplinkC6V4Router, TPLinkDecoClient, TplinkRouter]:
```
and test for your router TplinkRouterProvider.get_client(...) as there is a possibility that requests for TPLinkEXClient may take the main role before TPLinkMRClient. As we dont have now MR router - so it is better to test it in reverse order

AlexandrErohin: Would be great if you could add tests test_client_ex.py. Thank you!
In your client method supports calls self._get_url('cgi/getGDPRParm'). In MR client method supports calls self._get_url('cgi/getParm'), So your router can response on self._get_url('cgi/getGDPRParm') and self._get_url('cgi/getParm') with same response?

Adding routers model to the method supports - greatly reduces the possibilities. As I was very surprised how many different types routers have same API

krlosrenan: 1 - I added the tests (test_client_ex.py)
2 - Yes, the router I tested on responds to "cgi/getGDPRParm" and "cgi/getParm" in the same way.
3 - I believe this new model will support several other routers

Looking at the oid_str.js file, I noticed that there are possibly several other routers that may be supported.

Note the lines that start with INCLUDE_SPEC_

I will test it on a friend's EX220 soon.

I am attaching the file.
[oid_str.txt](https://github.com/user-attachments/files/17710044/oid_str.txt)

AlexandrErohin: 1 - Thank you for the tests
2 - Thats bad. Because this may break the MR support. I will try to find MR series router - to check response to "cgi/getGDPRParm". Hope it doesnt have

krlosrenan: I believe this won't be a problem because I adjusted the supports method so that it only responds positively for the EX511 and EX220 models. 

Before, the verification was only by requesting getGDPRParm, but now, in addition to that, the router needs to be on the list of supported models.

However, it would be great to test on a MR, because if there is no conflict, we could simplify the code so that it could support models that we haven't tested, but that use the same API.

If you need to change anything else, let me know.

AlexandrErohin:
Could you please post the hardware version for EX220 - when you test it?

AlexandrErohin:
I have tested it with MR router and it works - I have changed support method.
Could you pull the changes from main branch and test for your router please? If it works - Ill push the new client version

Thank you for help and for your work for adding support for your router

tplinkrouterc6u/client.py
```diff
@@ -1428,11 +1429,519 @@ def _prepare_data(self, data: str, is_login: bool) -> tuple[str, str]:
...
+        self._ee = None
+        self._seq = None
+
+        self._encryption = EncryptionWrapperMR()
+
+    def supports(self) -> bool:
+        try:
+            supported_routers = ['EX511', 'EX220']
+
+            url = self._get_url('js/oid_str.js')
```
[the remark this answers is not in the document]
krlosrenan: Note the lines that start with INCLUDE_SPEC_

I am attaching the file.
[oid_str.txt](https://github.com/user-attachments/files/17710044/oid_str.txt)

krlosrenan:
> I have tested it with MR router and it works - I have changed support method.
> Could you pull the changes from main branch and test for your router please? If it works - Ill push the new client version
> 
> Thank you for help and for your work for adding support for your router

Everything is working. I just did some tests and no errors occurred.

AlexandrErohin:
@krlosrenan Hi
Could you add
- get_vpn_status https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/mr.py#L684
- set_vpn https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/mr.py#L707
when you have time?

AlexandrErohin:
@lebdim Thank you for the info!
P.S. Would be great if you could give a star to the repo :) Thanks

PR #55, opened 2024-11-14
Add support for TP Link XDR3010

picone:
I've added support for the XDR3010 router. I tested in my network and everything work perfectly.

Hardware version: `TL-XDR3010易展版 2.0`
Software version: `1.0.18 Build 220711 Rel.56168n`


AlexandrErohin:
Thank you!
Could you implement all methods?

tplinkrouterc6u/client.py
```diff
@@ -1428,11 +1428,136 @@ def _prepare_data(self, data: str, is_login: bool) -> tuple[str, str]:
...
+class TPLinkXDR3010Client(AbstractRouter):
+    _stok = ''
+
+    def __init__(self, host: str, password: str, username: str = 'admin', logger: Logger = None,
+                 verify_ssl: bool = True, timeout: int = 30) -> None:
+        super().__init__(host, password, username, logger, verify_ssl, timeout)
+        self._session = Session()
+
+    def supports(self) -> bool:
+        response = self._session.get(self.host)
```
AlexandrErohin, line 1440: please dont forget to pass verify_ssl, timeout

tplinkrouterc6u/client.py
```diff
@@ -1428,11 +1428,136 @@ def _prepare_data(self, data: str, is_login: bool) -> tuple[str, str]:
...
+        return response.text.index('TL-XDR3010') >= 0
```
AlexandrErohin, line 1441: Is there a way to check supports not by checking the model? Adding routers model to the method supports - greatly reduces the possibilities. As I was very surprised how many different types routers have same API

tplinkrouterc6u/client.py
```diff
@@ -1428,11 +1428,136 @@ def _prepare_data(self, data: str, is_login: bool) -> tuple[str, str]:
...
+
+    def authorize(self) -> None:
+        response = self._session.post(self.host, json={
```
AlexandrErohin, line 1444: please dont forget to pass verify_ssl, timeout

tplinkrouterc6u/client.py
```diff
@@ -1428,11 +1428,136 @@ def _prepare_data(self, data: str, is_login: bool) -> tuple[str, str]:
...
+    def _request(self, payload: dict) -> dict:
+        url = '{}/stok={}/ds'.format(self.host, self._stok)
+        response = self._session.post(url, json=payload, timeout=self.timeout, verify=self._verify_ssl)
+        return response.json()
+
+    @staticmethod
+    def _encode_password(pwd: str) -> str:
+        return TPLinkXDR3010Client._security_encode(
+            pwd,
+            'RDpbLfCPsJZ7fiv',
```
AlexandrErohin, line 1532: Are you sure that these two strings are constant? Maybe there is a route to get them? Like this one https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client.py#L184

tplinkrouterc6u/client.py
picone: Or... try to login in? If it works, that means it's supported.

tplinkrouterc6u/client.py
picone: I'm not sure, because I don't think I have two XDR3010 routes.
This key is hardcoded in the JS file instead of being in the API, and I think TP Link's engineers wouldn't dynamically generate the JS file.

AlexandrErohin: If it in the JS file - agree

tplinkrouterc6u/client.py
```diff
@@ -1428,11 +1428,186 @@ def _prepare_data(self, data: str, is_login: bool) -> tuple[str, str]:
...
+            res += char_dict[(b ^ a) % dict_len]
+        return res
+
+
 class TplinkRouterProvider:
     @staticmethod
     def get_client(host: str, password: str, username: str = 'admin', logger: Logger = None,
                    verify_ssl: bool = True, timeout: int = 30) -> AbstractRouter:
-        for client in [TplinkC5400XRouter, TPLinkMRClient, TplinkC6V4Router, TPLinkDecoClient, TplinkRouter]:
+        for client in [TPLinkXDR3010Client, TplinkC5400XRouter, TPLinkMRClient, TplinkC6V4Router, TPLinkDecoClient,
```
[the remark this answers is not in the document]
picone: Get it, but I'm not sure how to write unit tests for the `supports()` methods of other routes just yet.

tplinkrouterc6u/client.py
picone: This might be a bit tricky; I used `Session().get()` directly for the request. Now, none of the other routes have been tested with this method.

tplinkrouterc6u/client.py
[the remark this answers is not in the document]
AlexandrErohin: So does get_client works for you when XDR3010 in the end for provider priority?

tplinkrouterc6u/client.py
AlexandrErohin: Ok. Could you just post here the output from print(response.text)? and change check to if model starts with string - TL-XDR ?

tplinkrouterc6u/client.py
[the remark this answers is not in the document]
picone: Yes, it works

tplinkrouterc6u/client.py
picone: The response is like to this:
```
<!DOCTYPE html>
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
	<title>TL-XDR3010易展版</title>
	<meta name="MobileOptimized" content="240" />
	<meta name="viewport" content="width=device-width, height=device-height, initial-scale=1.0, minimum-scale=0.5, maximum-scale=2.0, user-scalable=yes" />
	<link rel="shortcut Icon" href="../web-static/images/icon.ico" type="image/x-icon" />
	<link rel="stylesheet" href="../web-static/dynaform/class.css">
	<script type="text/javascript" src="../web-static/dynaform/class.js"></script>
	<script type="text/javascript" src="../web-static/dynaform/jtopo.js"></script>
</head>
<body>
	<div id="Error"></div>
	<div id="Confirm"></div>
	<div id="Con"></div>
	<div id="Help"></div>
	<div id="Cover"></div>
	<div id="Login"></div>
	<script type="text/javascript">
		var gBeInCNA = "NO";
		var proName="TL-XDR3010易展版";
		pageOnload();
    </script>
</body>
</html>
```

I'm not sure whether the routing for other XDR models is the same across the board. If it is, I should rename the class to `TPLinkXDRClient`, but I don't have other router models to test this on.

AlexandrErohin: Yes - It would be great to rename TPLinkXDRClient. After your good suggestion I have changed the files and now the file structure changed. Could you update your MR with changes from main branch and create tplinkrouterc6u/client/xdr.py and put the client to this file?
Also would be great if you could add methods
- get_ipv4_status https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/api_cgi_bin.py#L362C9-L362C24
- get_ipv4_reservations https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/api_cgi_bin.py#L380C9-L380C30
- get_ipv4_dhcp_leases https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/api_cgi_bin.py#L391C9-L391C29

AlexandrErohin:
Thank you!

picone:
> @picone Hi Could you add
> 
> * get_vpn_status https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/mr.py#L684
> * set_vpn https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/mr.py#L707
>   when you have time?

I checked my router settings and couldn't find any information about VPN functions.

AlexandrErohin:
Ok. Thats mean your router doesnt support VPN. Thank you

PR #56, opened 2024-11-19
Added VPN support on EX routers

krlosrenan:
Hi, I have added VPN support to the EX511


AlexandrErohin:
@krlosrenan Thank you very much!

PR #58, opened 2024-11-22
ex: allow specifying username by constructor

ktprograms:


AlexandrErohin:
@ktprograms Hi. You may pass different username to the constructor.

PR #67, opened 2025-02-24
Add support to TP-Link Archer VR1200

MatteoR23:
I've added support for the V1200 router. I tested in my network and everything work perfectly.

Firmware: 1.1.0 0.8.0 v600e.0 Build 231124 Rel.22957n 
Hardware: Archer VR1200v v1 00000000


MatteoR23:
@AlexandrErohin  hi, I made some review. I squashed all commits so you can better understand the changes I made. The code keeps working. You can merge it whenever you want.

tplinkrouterc6u/provider.py
```diff
@@ -10,13 +10,14 @@
...
 from tplinkrouterc6u.client.c1200 import TplinkC1200Router
+from tplinkrouterc6u.client.vr1200 import TPLinkVR1200Client
 
 
 class TplinkRouterProvider:
     @staticmethod
     def get_client(host: str, password: str, username: str = 'admin', logger: Logger = None,
                    verify_ssl: bool = True, timeout: int = 30) -> AbstractRouter:
-        for client in [TplinkC5400XRouter, TPLinkEXClient, TPLinkMRClient, TplinkC6V4Router, TPLinkDecoClient,
+        for client in [TPLinkVR1200Client, TplinkC5400XRouter, TPLinkEXClient, TPLinkMRClient, TplinkC6V4Router, TPLinkDecoClient,
```
AlexandrErohin, line 20: Unfortunately, It breaks MR compatibility. Your class has same implementation of method `supports` as TPLinkMRClient. So now MR routers fails as they got to your TPLinkVR1200Client. Is there a way to implement method `supports` that satisfy only routers that has same API with VR1200?

test/test_client_vr1200.py
```diff
@@ -0,0 +1,6 @@
+from test.test_client_mr import TestTPLinkMRClient
+
+
+class TestClientVR1200(TestTPLinkMRClient):
```
AlexandrErohin, line 4: This class tests TPLinkMRClient again as in TestTPLinkMRClient ` client = TPLinkMRClient('', '')` https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/test/test_client_mr.py#L58

AlexandrErohin:
@MatteoR23 Hi. Thank you for adding Archer VR1200v v1 support!
I have added comments. Could you check it please? It breaks MR compatibility :(

tplinkrouterc6u/provider.py
MatteoR23: @AlexandrErohin 

Not really, because supports checks if it can generate rsa.
My implementation uses a different endpoint.

See:
self._url_rsa_key = 'cgi/getGDPRParm'

```python
    class TPLinkVRClientBase(TPLinkMRClientBase):

    def __init__(self, host: str, password: str, username: str = 'admin', logger: Logger = None,
                 verify_ssl: bool = True, timeout: int = 30):
        super().__init__(host, password, username, logger, verify_ssl, timeout) 
        self._url_rsa_key = 'cgi/getGDPRParm'
```

test/test_client_vr1200.py
MatteoR23: Yes, as it uses the same test structure.
I would copy/paste it, but it's not something I would like to do...

tplinkrouterc6u/provider.py
AlexandrErohin: Oh, I was wrong. It breaks EX compatibility TPLinkEXClient as it has also self._url_rsa_key = 'cgi/getGDPRParm'. Is there any other way to determine VR1200 API?

test/test_client_vr1200.py
AlexandrErohin: Agree

MatteoR23:
@AlexandrErohin  it works now :-)

MatteoR23:
@AlexandrErohin  it works now :-)

tplinkrouterc6u/client/vr1200.py
```diff
@@ -0,0 +1,163 @@
...
+
+class TPLinkVRClientBase(TPLinkMRClientBase):
+
+    def __init__(self, host: str, password: str, username: str = 'admin', logger: Logger = None,
+                 verify_ssl: bool = True, timeout: int = 30):
+        super().__init__(host, password, username, logger, verify_ssl, timeout) 
+        self._url_rsa_key = 'cgi/getGDPRParm'
+
+    def supports(self):
+        return self._verify_modem() and self._req_rsa_key()
```
AlexandrErohin, line 19: Could we try something more general then identifying of a specific model? I think there are some models with same API. And make code work only for specific model limits the client.

MatteoR23: @AlexandrErohin 

All right, I have a proposal:
I could check if this is an Archer VR1*** series, I suspect that all of them works on the same way.
I would avoid to extend this to all VR*** because 3 of  them are fully tested and working and I would not touch them.

On this way, I would not break other functioning router.

Do you agree this approach?

MatteoR23: @AlexandrErohin I enhanced the support method. Please, have a look and let me know :-)
Basically, I do some things: 

- I test on the router web page if there's "Archer VR" on it
- I test if on the router web page or lib.js if the cgi login is compatible
- I test three CGI calls:
- I tested some API to check compatibility:
     - authorize
     - status
     - logout

These checks should be enough to check API compatibility, at least most of them.
Maybe, we can think to extend this approach (totally or partially) to the other routers.

What do you think?

AlexandrErohin: @MatteoR23 I checked other API and EX also has all this endpoints. Testing authorization in `support` method leads to not supported if password mistyped. Maybe lets try that you suggested before only check if router model starts VR and and self._req_rsa_key() ?

MatteoR23: @AlexandrErohin  Sorry for the late reply, I had personal issues... 

It's okay for me :-) The problem here is that EX is able to generate RSA data, but there are some differences on the way the router is consuming those endpoints (even if the name suggests that are the same, they are not behind them).
For example:
-  the login requires that the password is encoded in base64 before AES encoding.
- Logout has a different return type, that is totally incompatible with EX.
- The other endpoints works as expected

So, the only way to be 100% sure that the API are supported is to test them.

MatteoR23:
@AlexandrErohin allright, now should be much better :-)
Thank you for your suggestions

MatteoR23:
@AlexandrErohin  I fixed all flake8 errors and warnings

AlexandrErohin:
@MatteoR23 Thank you! I have merged it

PR #68, opened 2025-03-03
TPLinkVRClient update

AlexandrErohin:


AlexandrErohin:
@MatteoR23 Hi. I have deleted some duplicate code and unused parameters. Could you test it please?

tplinkrouterc6u/client/vr.py
```diff
@@ -132,83 +107,23 @@ def _req_login(self) -> None:
...
-        else:
-            return r.status_code, r.text
-
-
-class TPLinkVRClient(TPLinkVRClientBase, TPLinkMRClient):
-    def __init__(self, host, username, password, logger=None, verify_ssl=True, timeout=30):
-        super().__init__(host, username, password, logger, verify_ssl, timeout)
+    def _request(self, url, method='POST', data_str=None, encrypt=False, is_login=False):
+        is_login = encrypt and '/cgi/login' in data_str
+        super()._request(url, method, data_str, encrypt, is_login)
```
MatteoR23, line 129: @AlexandrErohin  It doesn't work.
Row number 129 should be:
**return** super()._request(url, method, data_str, encrypt, is_login)

Add "return", that is missing, as code expects a tuple

MatteoR23 (changes requested):
@AlexandrErohin  It seems that it's not working, I suggested a change

tplinkrouterc6u/client/vr.py
AlexandrErohin: @MatteoR23 Thanks! You are right! I have fixed it. All methods are work ok? login, logout, get_firmware, get_status, get_ipv4_reservations, get_ipv4_dhcp_leases, get_ipv4_status, set_wifi, send_sms, get_sms, set_sms_read, delete_sms, send_ussd, get_lte_status, get_vpn_status, set_vpn ?

MatteoR23: @AlexandrErohin 

The following are fully tested:

login, logout, get_firmware, get_status, get_ipv4_reservations, get_ipv4_dhcp_leases, get_ipv4_status, set_wifi, get_vpn_status, set_vpn 

What's not tested:

- My router can't send SMS or get LTE status :-(  I mean, it can do it, but it needs a 4G USB device that I don't have In this case I get an error from the API, but it's something expected.

AlexandrErohin: @MatteoR23 Thanks!
Could you post the error please for methods send_sms, get_sms, set_sms_read, delete_sms, send_ussd, get_lte_status ?

MatteoR23: @AlexandrErohin 

I tested them:
- get_sms returns always a 0 length array. No errors risen from mr.py
- send_sms simply do nothing. No errors risen from mr.py
- get_lte_status is broken, since it expects that "values" has at least one item, but in this case is 0 length and an index out of range error. 
- Same story for send_ussd.
- set_sms_read, delete_sms cannot be tested at all, since I don't have an SMS to read or to delete

for them, the API sends this:

status_code: 200
response: [error]9804

I looked at router's js, I found a constant called CMM_OBJECT_NOT_FOUND equal to 9804.

AlexandrErohin: @MatteoR23 Thank you! I will finally test it in HA and release HA integration in a few days

MatteoR23: @AlexandrErohin  Great!! Thank you for your patience!!

PR #85, opened 2025-06-09
Added exception handling the ex.py client as some routers return 406 on the Memory request

FriedrichM:
As described in https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/180 sometimes certain routers return 406  for the memory state request.
This is one way to handle those 406 responses feel free to give me feedback on this draft. 
Do you want to handle it in another way ?

### DIFF
--- a/tplinkrouterc6u/client/ex.py
+++ b/tplinkrouterc6u/client/ex.py
@@ -74,7 +74,7 @@ class TPLinkEXClient(TPLinkMRClientBase):
         ]
         _, values = self.req_act(acts)
 
-        if not values:
+        if not values or values[0] is None:
             raise ValueError('No firmware information received.')
 
         firmware = Firmware(
@@ -99,6 +99,9 @@ class TPLinkEXClient(TPLinkMRClientBase):
         ]
 
         _, values = self.req_act(acts)
+
+        if not values or any(x is None for x in values[:-2]):
+            raise ValueError("Incomplete or malformed status data from device.")
 
         if values[0].__class__ == list:
             values[0] = values[0][0]
@@ -143,11 +146,17 @@ class TPLinkEXClient(TPLinkMRClientBase):
                                                  IPv4Address(val['IPAddress']),
                                                  val['hostName'])
 
-        total = int(values[4]['total'])
-        free = int(values[4]["free"])
-        status.mem_usage = ((total - free) / total)
+        try:
+            total = int(values[4]['total'])
+            free = int(values[4]["free"])
+            status.mem_usage = ((total - free) / total)
+        except ValueError:
+            status.mem_usage = 0
 
-        status.cpu_usage = int(values[5]['CPUUsage']) / 100
+        try:
+            status.cpu_usage =int(values[5]['CPUUsage']) / 100
+        except ValueError:
+            status.cpu_usage = 0
 
         status.devices = list(devices.values())
         status.clients_total = status.wired_total + status.wifi_clients_total + status.guest_clients_total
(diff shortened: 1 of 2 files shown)


FriedrichM:
ive been able to recreate the 406 error code by running the readme script in quick succession / even in parallel. so it seems to be tied to that maybe something with logins or request ratelimiting

AlexandrErohin:
Hi!
Thank you for your PR!
If you would like to add retries - it is better to add it to client directly - like https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/mr.py#L543

tplinkrouterc6u/client/ex.py
```diff
@@ -100,6 +100,9 @@ def get_status(self) -> Status:
 
         _, values = self.req_act(acts)
 
+        if not values or any(x is None for x in values[:-2]):
```
AlexandrErohin, line 103: Would be great if you could add test for this behavior

tplinkrouterc6u/client/ex.py
```diff
@@ -143,11 +146,17 @@ def get_status(self) -> Status:
                                                  IPv4Address(val['IPAddress']),
                                                  val['hostName'])
 
-        total = int(values[4]['total'])
-        free = int(values[4]["free"])
-        status.mem_usage = ((total - free) / total)
+        try:
+            total = int(values[4]['total'])
```
AlexandrErohin, line 150: Would be great if you could add test for this behavior

tplinkrouterc6u/client/ex.py
```diff
@@ -143,11 +146,17 @@ def get_status(self) -> Status:
...
+            free = int(values[4]["free"])
+            status.mem_usage = ((total - free) / total)
+        except Exception:
+            status.mem_usage = 0
 
-        status.cpu_usage = int(values[5]['CPUUsage']) / 100
+        try:
```
AlexandrErohin, line 156: Would be great if you could add test for this behavior

tplinkrouterc6u/client/ex.py
```diff
@@ -178,6 +189,8 @@ def get_ipv4_dhcp_leases(self) -> [IPv4DHCPLease]:
                          attrs=['IPAddress', 'physAddress', 'hostName', 'leaseTimeRemaining']),
         ]
         _, values = self.req_act(acts)
+        if not values or values[0] is None:
```
AlexandrErohin, line 192: Would be great if you could add test for this behavior

tplinkrouterc6u/client/ex.py
```diff
@@ -202,6 +215,8 @@ def get_ipv4_status(self) -> IPv4Status:
                                 'connIPv4DnsServer']),
         ]
         _, values = self.req_act(acts)
+        if not values or any(v is None for v in values):
```
AlexandrErohin, line 218: Would be great if you could add test for this behavior

tplinkrouterc6u/client/ex.py
```diff
@@ -74,7 +74,7 @@ def get_firmware(self) -> Firmware:
         ]
         _, values = self.req_act(acts)
 
-        if not values:
+        if not values or values[0] is None:
```
AlexandrErohin, line 77: Would be great if you could add test for this behavior

tplinkrouterc6u/client/ex.py
```diff
@@ -159,6 +168,8 @@ def get_ipv4_reservations(self) -> [IPv4Reservation]:
             self.ActItem(self.ActItem.GL, 'DEV2_DHCPV4_POOL_STATICADDR', attrs=['enable', 'chaddr', 'yiaddr']),
         ]
         _, values = self.req_act(acts)
+        if not values or values[0] is None:
```
AlexandrErohin, line 171: Would be great if you could add test for this behavior

PR #101, opened 2025-10-24
Update mr200.py

potat0man:


AlexandrErohin:
@potat0man Could you run this code please?
```
from tplinkrouterc6u import TplinkRouterProvider

router = TplinkRouterProvider.get_client('http://192.168.0.1', 'your_plain_password')
print(router.get_firmware().__dict__)
print(router.get_status().__dict__)
```

PR #105, opened 2025-11-07
Feature/c80/RE330 aes init

groberth:
Hello again 👋 

I switched AES encryption plugin instead of copy/pasting the code.
The main goal was to upgrade the AES key initialisation instead of using the default one.

Flake Ok
Tests Ok

Tested on RE330 but not on C80 (which are very similar, should be okay)


AlexandrErohin:
@groberth Thank you!

PR #107, opened 2025-11-12
Update c6u.py

david9600:
Store main wifi client count (obtained from admin/status), so that total wifi count can be incremented by any new devices in the statistics array (for routers which support it). 
Also set the IoT client count (if not already set by smart_network) by subtracting main and guest from total wifi.


tplinkrouterc6u/client/c6u.py
```diff
@@ -360,6 +360,9 @@ def _add_device(conn: Connection, item: dict) -> None:
                 devices[item['mac']].up_speed = item.get('uploadSpeed')
                 devices[item['mac']].signal = int(item.get('signal')) if item.get('signal') else None
 
+        # Store main wifi count (so only the total will be incremented for any new devices from wireless statistics array)
+        status.main_wifi_clients_total = status.wifi_clients_total
```
AlexandrErohin: It is better not to add new counters, because there are also other clients that dont have it.

Issue #112, opened 2025-11-15
Device active always returns "True"


mikeluyten:
https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/b590efa023c7577de1b02add2cf864f493ec288f/tplinkrouterc6u/common/dataclass.py#L26

I have a BE220/BE3600 that uses the c6u.py client, but active is never set to True/False in c6u.py so I beleive it defaults back to the dataclass.py definition of a device "True", even for inactive clients.

dennismdejong:
> *This was generated by AI during triage.*

## Triage Notes

**What we have established so far:**
- The `Device` dataclass defaults `active: bool = True` (`dataclass.py:33`), so any device entry parsed without an explicit flag is reported as active.
- In `c6u.py` the `_add_device` helper parses `access_devices_wired` / `access_devices_wireless_host` / `access_devices_wireless_guest` and never reads an "active" field — it only takes `macaddr`, `ipaddr`, `hostname`.
- The three PRs that referenced this issue (#110, #113, #137) were closed without merging and none of them touched the active field, so it remains unimplemented.

**What we still need from you (@mikeluyten):**
- The raw JSON from the status call so we can see which field marks a client inactive. Please run:
  ```python
  router = TplinkRouterProvider.get_client("http://192.168.0.1", "password")
  router.authorize()
  import json; print(json.dumps(router.request("admin/status?form=all&operation=read", "operation=read"), indent=2))
  ```
  and share the output (mask any sensitive values). We specifically need one device entry that is **active** and one that is **inactive** so we can identify the distinguishing field.
- Which exact router model + firmware version are you on?

PR #113, opened 2025-11-17
Adding Guest wifi management and extra device properties

mikeluyten:
#112 
Appologies for the new PR, had to clear out some bad commit history

### DIFF
--- a/tplinkrouterc6u/client/c6u.py
+++ b/tplinkrouterc6u/client/c6u.py
@@ -256,7 +256,7 @@ class TplinkBaseRouter(AbstractRouter, TplinkRequest):
     def set_wifi(self, wifi: Connection, enable: bool = None, ssid: str = None, hidden: str = None,
                  encryption: str = None, psk_version: str = None, psk_cipher: str = None, psk_key: str = None,
                  hwmode: str = None, htmode: str = None, channel: int = None, txpower: str = None,
-                 disabled_all: str = None) -> None:
+                 disabled_all: str = None, portal_password: str = None) -> None:
         values = {
             Connection.HOST_2G: 'wireless_2g',
             Connection.HOST_5G: 'wireless_5g',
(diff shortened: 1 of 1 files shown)


tplinkrouterc6u/client/c6u.py
```diff
@@ -264,11 +268,78 @@ def set_wifi(self, wifi: Connection, enable: bool) -> None:
...
+        guest_wifi.guest_2g_portal_password = data.get('guest_2g_portal_password')
+        guest_wifi.guest_5g_enable = self._str2bool(data.get('guest_5g_enable'))
+        guest_wifi.guest_5g_ssid = data.get('guest_5g_ssid')
+        guest_wifi.guest_5g_encryption = data.get('guest_5g_encryption')
+        guest_wifi.guest_5g_psk_key = data.get('guest_5g_psk_key')
+        guest_wifi.guest_5g_portal_enable = self._str2bool(data.get('guest_5g_portal_enable'))
+        guest_wifi.guest_5g_portal_password = data.get('guest_5g_portal_password')
+        return guest_wifi
+
+    def set_guest_wifi_password(self, password: str, band: str = '2g5g') -> None:
```
AlexandrErohin: set_wifi already do it - no need to add set_guest_wifi_password

test/test_client_c6u.py
```diff
@@ -10,10 +10,50 @@
...
+        self.assertEqual(guest_wifi_info.guest_2g_ssid, "guest_2g")
+        self.assertEqual(guest_wifi_info.guest_2g_encryption, "psk")
+        self.assertEqual(guest_wifi_info.guest_2g_psk_key, "password")
+        self.assertEqual(guest_wifi_info.guest_2g_portal_enable, False)
+        self.assertEqual(guest_wifi_info.guest_5g_enable, False)
+        self.assertEqual(guest_wifi_info.guest_5g_ssid, "guest_5g")
+        self.assertEqual(guest_wifi_info.guest_5g_encryption, "none")
+        self.assertEqual(guest_wifi_info.guest_5g_psk_key, "")
+
     def test_get_status(self) -> None:
```
AlexandrErohin: it is better to add to old test only checks that your new fields are None for case when router doesnt have these fields. And add a new test that checks all fields values with your new fields.

PR #115, opened 2025-11-17
Enhance set_wifi method with more configuration options

mikeluyten:
Updated set_wifi method to accept additional parameters for configuration.


mikeluyten:
Is this more what you were after @AlexandrErohin? I think I'm going to do this bit by bit, I'm having some serious github issues, having to delete my fork daily.

PR #120, opened 2025-12-05
add support for VR400 v2

EldarZ:
Added support for tplink VR400 v2, based on MRClient.
Validated that the example script is functional.

hardware_version='Archer VR400 v2 00000000'
model='Archer VR400'
firmware_version='0.2.0 0.9.1 v0070.0 Build 170719 Rel.63031n'


lostonline22:
Added support for tplink Archer BE450 v1.0

AlexandrErohin:
@EldarZ Hi! It looks like authorize is the same with TPLinkMR200Client.
And clients order in provider.py breaks MR routers support - the new class should be added after TPLinkMR200Client

PR #124, opened 2026-01-01
Support for Archer 3200

HaJJ-MBX:
Support for Archer 3200 router.

Two main differences with other clients :
- The C3200 router needs a base64 encoded cookie to authorize. It is implemented in the support method.
- The CGI logic comes mainly from MR Clients but was not directly compatible. I have rewritten it, it could be mutualized but I have not investigated.

### DIFF
--- a/test/test_client_C3200.py
+++ b/test/test_client_C3200.py
@@ -7,9 +7,7 @@ from macaddress import EUI48
 from tplinkrouterc6u import (
     TplinkC3200Router,
     Connection,
-    Firmware,
     Status,
-    Device,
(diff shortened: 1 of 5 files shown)


HaJJ-MBX:
yes there is a vpn within the router, never tested but I can implement it. 
I'll have a look at extending the MR200 client. In the mean time I am eager to see it live on my HA Green !

AlexandrErohin:
TPLinkMR200Client already has get_vpn_status and set_vpn - so it would be great if you could test to use it with override supports, authorize, _request

PR #126, opened 2026-01-13
Add support for TL-R470GP-AC and TL-R488GPM-AC

0x5e:
Tested on:
TL-R470GP-AC 4.0, 2.0.0 Build 211008 Rel.48276n
TL-R488GPM-AC 2.0, 2.0.0 Build 220304 Rel.12108n
Both are AC.

The TL-R series use `stok` for login, similar to XDR series.
It's model name didn't show up in `/` html title, have to get it on `/login.htm`.
`username` is required in authorize method.
Most of the api is similar, different in key names.

And the test case is included, only one part is broken:
I already add `unquote` in `TPLinkRClient`'s parent `TPLinkXDRClient` to fix the quoted hardware version, but the testcase always failed, do you know how to fix this?

### DIFF
--- a/test/test_client_r.py
+++ b/test/test_client_r.py
@@ -78,24 +78,6 @@ $(document).ready(function(e){
 
         self.assertEqual(client.supports(), True)
 
-    def test_logout(self) -> None:
-        mock_data = json.loads('''{"error_code":0}''')
-        check_payload = {}
-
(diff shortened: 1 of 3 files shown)


AlexandrErohin:
Unfortunately, I don't have XDR series router - so I cannot test authorize method. It is better to add authorize method to your class so as not to break XDR compatibility

PR #136, opened 2026-01-30
feat: add support for SG_L1_S2 / CE_RED certified routers (SHA256 + OAEP)

mattmanning:
## Summary

Adds support for TP-Link routers with **SG CLS L1 STAGE2** or **EU CE RED** certification (e.g., Archer BE3600 and other Wi-Fi 7 models). These routers use an enhanced encryption scheme that differs from the standard MD5 + PKCS1_v1_5 approach.

## Problem

After firmware updates, routers with these certifications fail to authorize with:
```
TplinkRouter - Cannot authorize! Error - Expecting value: line 1 column 1 (char 0); Response -
```

This is because the router returns HTTP 403 when the standard encryption scheme is used.

Fixes https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/220

## Root Cause

Routers with SG_L1_S2 / CE_RED certifications use a different encryption scheme, determined by the `certification` field in `/device_config?form=config`:

| Feature | Standard | SG_L1_S2 / CE_RED |
|---------|----------|-------------------|
| Auth hash | MD5(`username` + `password`) | **SHA256**(`username` + `password`) |
| Login signature RSA | PKCS1 v1.5 | **PKCS1-OAEP** |
| Request signature | RSA-encrypted chunks | **HMAC-SHA256** chunks |
| Hash per request | Static | **SHA256(encrypted_data)** |

## Changes

- New `tplinkrouterc6u/client/sg.py` — `TplinkRouterSG` client class
- Auto-detects SG/CE_RED routers via certification check
- Added to provider client list (before `TplinkRouterV1_11`)
- Exported from `__init__.py`
- Auth hash uses `self.username` (not hardcoded `'admin'`)
- Removed redundant `import hashlib`; use already-imported `sha256` in HMAC call
- Removed dead-code `_prepare_data`/`_decrypt_response` overrides
- New `test/test_client_sg.py` with 27 tests (auth flow, certification detection, HMAC signatures, inherited behavior)

## Test plan

- [x] `pytest test/test_client_sg.py -v` — all 27 tests pass
- [x] `pytest test/ -v` — full suite of 185 tests pass
- [x] `flake8 tplinkrouterc6u/client/sg.py test/test_client_sg.py` — clean

## Tested Against

```
Hardware: Archer BE3600 v1.0
Firmware: 1.2.1 Build 20250804 rel.79639(5553)
Certifications: US FCC, SG CLS L1 STAGE2
```

Output:
```
Client type: TplinkRouterSG
Router: Archer BE3600 Archer BE3600 v1.0
Firmware: 1.2.1 Build 20250804 rel.79639(5553)
Clients: 36 (6 wired, 30 wifi)
```

### DIFF
--- a/tplinkrouterc6u/client/sg.py
+++ b/tplinkrouterc6u/client/sg.py
@@ -66,8 +66,6 @@ class TplinkRouterSG(TplinkBaseRouter):
             if not self._check_sg_certification():
                 return False
             self._request_pwd_keys()
-            self.authorize()
-            self.logout()
             return True
         except Exception:
             return False


mattmanning:
Tested all three detection methods on my BE3600 — `ui-type=svr`, keys fallback, and certification check all succeed. These are the same router type.

| Detection method | BE3600 result |
|---|---|
| `ui-type=svr` meta tag (#137) | ✅ |
| Keys fallback: mode=router, username='', key≥512 (#137) | ✅ |
| SG certification (#136) | ✅ |

Happy to close this in favor of #137. @hepter can also incorporate the test coverage from this PR if they'd like, or I can rework this PR to adopt #137's detection and retry logic. Please let me know how I can help. Cheers and thanks for this project.

hepter:
Thanks for testing all three detection paths 🙌 @mattmanning 
I agree #136 is the cleaner and more targeted base to continue from.

Just a heads‑up from my BE230: /login?form=keys returns username="", while HA passes username=admin by default. So even if detection is correct, there’s a chance the hash could be computed with admin and fail on some BE devices expecting an empty username. I’m mentioning this only as a possible edge case to keep in mind.

Thanks again for the work and tests.
I verified your approach on BE230 and it matches my router.

PR #138, opened 2026-02-01
Add extra per-client metrics from smart_network

hepter:
## Summary
  - Add per-client tx/rx rate, online time and traffic usage from smart_network `loadDevice`.
  - Merge by MAC and keep existing data when the endpoint is missing.
  - Expose these values on the Device dataclass for downstream consumers.
  - Added tests: `test_get_status_with_game_accelerator_fallback_values`, `test_get_status_with_game_accelerator_alt_keys`.

  ## Tested
  - Archer BE230 v1.0
    - Firmware Version: 1.2.5 Build 20260108 rel.68570(4555)

  ## Tests
  - `pytest test -v` - all 164 tests pass
  - `flake8` - clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,10 @@ or you have TP-link C5400X or similar router you need to get web encrypted passw
 | packets_received | total packets received | int, None |
 | down_speed | download speed | int, None |
 | up_speed | upload speed | int, None |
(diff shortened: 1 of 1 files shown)


hepter:
Thanks for the review. I’ve added tests to cover the new smart_network metrics.

PR #155, opened 2026-04-26
xdr: support new firmware MD5 nonce login (TL-7DR7270 1.0.18+)

xyb:
## Summary

- Newer TP-Link XDR/7DR firmware (e.g. TL-7DR7270 1.0.18+) rejects the legacy `orgAuthPwd` (XOR `_security_encode`) login. The new flow probes `user_management.get_encrypt_info` for a server nonce + supported `encrypt_type` list, then logs in with `MD5(password + ":" + nonce).hex()` and `encrypt_type: "3"`.
- `TPLinkXDRClient.authorize()` now probes first; if the response contains `encrypt_type` with `"3"` and a non-empty `nonce`, the new MD5 path is used. Otherwise it falls back to the original `_security_encode` payload byte-for-byte, so existing TL-XDR / TL-7DR6430 deployments keep working.
- Probe uses `min(self.timeout, 5)` for numeric timeouts (faster fail when probing legacy firmware) and passes tuple/None timeouts through unchanged. Probe failures (`ValueError` / `RequestException`) fall back to legacy.

## Test plan

- [x] Tested live against TL-7DR7270 1.0.18 (new firmware, MD5 path)
- [x] All 26 ``test_client_xdr.py`` unit tests pass, including:
  - ``test_authorize_new_firmware_md5`` - new MD5 path with ``encrypt_type`` containing ``"3"``
  - ``test_authorize_legacy_when_encrypt_type_lacks_3`` - legacy fallback when probe lacks ``"3"``, asserts exact body + key order
  - ``test_authorize_falls_back_on_*`` - fallbacks for missing/non-string nonce, non-zero ``error_code``, ``RequestException``, non-iterable ``encrypt_type``
  - ``test_authorize_normalizes_*_encrypt_type`` - list/tuple/set/scalar/integer normalization
  - ``test_authorize_*_timeout`` - ``min(numeric, 5)`` cap and tuple/None pass-through

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,6 +408,7 @@ or you have TP-link C5400X or similar router you need to get web encrypted passw
 - RE330 v1
 - TD-W9960 (v1, V1.20)
 - TL-7DR6430 1.0
+- TL-7DR7270 1.0 (firmware 1.0.18+, new auth)
 - TL-MR100 v2.0
 - TL-MR100-Outdoor v1.0
 - TL-MR105
--- a/test/test_client_xdr.py
+++ b/test/test_client_xdr.py
@@ -631,5 +631,289 @@ maximum-scale=2.0, user-scalable=yes" />
         self.assertEqual(ipv4_status.lan_ipv4_netmask_address, get_ip('0.0.0.0'))
 
 
+    def test_authorize_new_firmware_md5(self) -> None:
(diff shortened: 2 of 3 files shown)


AlexandrErohin:
@xyb Thank you!

PR #156, opened 2026-04-29
Handle empty response data without crashing

tjriley33:
## Problem

Some TP-Link firmwares — notably newer Deco builds — respond to certain admin endpoints (e.g. `admin/wireless?form=wlan`) with `{"data": ""}` instead of an encrypted payload. Today that fails in two unhelpful steps:

1. `EncryptionWrapper._unpad` raises:
   ```
   TypeError: ord() expected a character, but string of length 0 found
   ```
   because `s[len(s) - 1:]` is `''` on empty input.
2. `TplinkEncryption._decrypt_response` then surfaces that as an opaque `ClientError`, which kills the entire status update for callers. The visible victim is the Home Assistant `tplink_router` integration — every poll throws and no router data updates.

Reproduces reliably on a TP-Link Deco mesh on recent firmware: the wireless config endpoint responds with empty data while the rest of the endpoints work fine.

## Fix

- `EncryptionWrapper._unpad` returns the empty value untouched. No-op for any non-empty input — same code path as before.
- `TplinkEncryption._decrypt_response` returns `{}` when `data['data']` is empty or missing, letting the caller fall through to its standard "invalid response" handling instead of bubbling up the underlying decode error.

Net effect: status calls succeed for all endpoints that *do* return data; affected sensors (e.g. `wifi_*_enable` for Decos that don't implement the wireless endpoint) report unknown instead of crashing the whole update.

## Tests

Adds:
- `test/test_common_encryption.py` — empty-string and empty-bytes `_unpad` inputs, plus an AES round-trip sanity check.
- `test/test_client_c6u.py::test_decrypt_response_empty_data_returns_empty_dict` — empty `{'data': ''}` and missing-key cases.

Full suite passes locally (292 tests).

### DIFF
--- a/test/test_client_c6u.py
+++ b/test/test_client_c6u.py
@@ -1321,6 +1321,20 @@ class TestTPLinkClient(TestCase):
         with self.assertRaises(ClientException):
             client.get_vpn_client_status()
 
+    def test_decrypt_response_empty_data_returns_empty_dict(self) -> None:
+        """Some firmwares (e.g. newer Deco) return {'data': ''} for endpoints they
+        do not implement. _decrypt_response must not raise on those — return {} so the
+        caller falls through to its standard 'invalid response' handling instead of
+        bubbling up an opaque ord()/JSON error."""
+        from tplinkrouterc6u.client.c6u import TplinkEncryption
+
+        class _Stub(TplinkEncryption):
+            def __init__(self):
+                pass
+
+        self.assertEqual(_Stub()._decrypt_response({'data': ''}), {})
+        self.assertEqual(_Stub()._decrypt_response({}), {})
+
 
 if __name__ == '__main__':
     main()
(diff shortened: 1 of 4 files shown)


AlexandrErohin:
@tjriley33 Hi. Thank you for the MR. What is your router hardware version which you have used for tests and what methods have you tested?

AlexandrErohin:
@tjriley33 Hi. Any news?

AlexandrErohin:
@tjriley33 Any news?

tjriley33:
Hey sorry for the delay, i've been out of town. BE1100's were used to test it. Two unit tests and then my testing on my local system. It's totally cool if you want to cancel it. Love your integration! Appreciate all the effort.

AlexandrErohin:
@tjriley33 Thank you!

PR #161, opened 2026-06-03
fix: get signal level and more lte/5g info for some mobile routers (nx200)

kasulio:
Hello, for my Archer NX200 (v2.0, running firmware 1.3.0) the LTE/5g signal level and other indicators like rsrp, rsrq and snr were 0.

i saw that this was mentioned for another router (NE200) as issues for the ha integration:
https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/336
https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/318
i suppose my changes should fix it for this router as well.

the information about the signal level is behind another query (DEV2_LTE_SERVING_CELL_INFO), split up for different access technologies (5G, LTE, 3G, etc.)

i adjusted the code in a way that keeps the way it was before (extraction from DEV2_LTE_NET_STATUS) as a fallback. i'm not exactly sure if this is the correct way to do it, or maybe there should be a new subclass for 5g routers. can you give me some guidance on this @AlexandrErohin?

happy to adjust as needed

PR #162, opened 2026-06-03
add sms sending capabilities for nx200

kasulio:
hello,

i added sms sending capabilities for the nx200 router. it should probably work for some other models too.

it should fix this issue:
https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/201

i just submitted another pr (#161) where i asked whether it should live in the ex.py file or we should make a subclass for these 5g routers. i was wondering about the same question for this pr.

thanks for maintaining this project, it's great!

### DIFF
--- a/tplinkrouterc6u/client/ex.py
+++ b/tplinkrouterc6u/client/ex.py
@@ -269,6 +269,17 @@ class TPLinkEXClient(TPLinkMRClientBase):
 
         return status
 
+    def send_sms(self, phone_number: str, message: str) -> None:
+        acts = [
+            self.ActItem(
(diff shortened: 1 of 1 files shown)


kasulio:
tested it through home assistant, seems to work!

AlexandrErohin:
@kasulio Hi! Thank you! Could you also add test for new method to https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/test/test_client_ex.py

AlexandrErohin:
@kasulio Thank you! Could you add also get_sms for your router model?

PR #163, opened 2026-06-05
add sms retrieving capabilities for nx200

kasulio:
hello, i implemented a few more methods for sms. now you should be able to get in the inbox as a list and mark single ones as read or delete them.

pagination behaves a bit tricky and to select a specific sms you have to provide a page and stack number. because this is not stable (for example after deleting one) it is necessary to always get the newest state of the sms list to do it beforehand. this may result in a lot of fetch calls, when doing bulk "mark as read" or "delete" operations.

### DIFF
--- a/test/test_client_ex.py
+++ b/test/test_client_ex.py
@@ -1,4 +1,5 @@
 from unittest import main, TestCase
+from datetime import datetime
 from macaddress import EUI48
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
Thank you!

PR #166, opened 2026-06-16
Fix TPLinkRClient handling of offline clients without IP addresses

rz467fzs7d:
## Summary

- Allow `TPLinkRClient.get_status()` to parse offline host entries that do not include an `ip` field.
- Preserve TP-Link AC/AP metadata on `Device`: `ap_name`, `ssid`, and `frequency`.
- Prefer `rssi` as the signal source for TP-Link R/AC host entries while keeping `signal` as a fallback.

## Why

Some TP-Link R/AC firmwares return offline entries in `host_management.host_info` with a MAC, hostname, type, and state, but without `ip`. The previous parser raised `KeyError: ip`, which prevents Home Assistant integrations from setting up even though most clients are readable.

## Validation

- `python3 -m unittest test.test_client_r.TestTPLinkRClient.test_get_status -v`
- `python3 -m unittest discover -s test -v`

Both pass locally; full suite: 314 tests OK.

### DIFF
--- a/tplinkrouterc6u/client/r.py
+++ b/tplinkrouterc6u/client/r.py
@@ -98,15 +98,24 @@ class TPLinkRClient(TPLinkXDRClient):
                 conn_type = Connection.HOST_5G
                 status.wifi_clients_total += 1
 
-            dev = Device(conn_type, get_mac(item['mac']), get_ip(item['ip']), unquote(item['hostname']))
+            ipaddr = get_ip(item['ip']) if item.get('ip') else None
+            dev = Device(conn_type, get_mac(item['mac']), ipaddr, unquote(item['hostname']))
             if 'up_speed' in item:
                 dev.up_speed = int(item['up_speed'])
             if 'down_speed' in item:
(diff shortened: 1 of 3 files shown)


AlexandrErohin:
@rz467fzs7d Tyhank you! What is your router hardware version?

AlexandrErohin:
@rz467fzs7d THi! What is your router hardware version?

tplinkrouterc6u/client/r.py
```diff
@@ -98,15 +98,24 @@ def get_status(self) -> Status:
                 conn_type = Connection.HOST_5G
                 status.wifi_clients_total += 1
 
-            dev = Device(conn_type, get_mac(item['mac']), get_ip(item['ip']), unquote(item['hostname']))
```
AlexandrErohin, line 101: I think it is better to use like this
```
Device(conn_type, get_mac(item['mac']), get_ip(item.get('ip')), unquote(item['hostname']))
```
If is no ip - it would be default 0.0.0.0 rather then None. it is better for using when IP is required and cannot be None

PR #172, opened 2026-07-19
Add WAN IPv6 support for MR client (tested on VR1600v v1, f/w v2.1.0)

david9600:
This adds IPv6 WAN support for routers which use MR Client.
A new method, get_ipv6_status, is added to MR Client Base, returning several fields.
The standard get_status method has just one new field - the WAN IPv6 address.  I found it a useful indication of whether the IPv6 connection is active on a dual stack ISP service.
Both methods have been tested successfully using an Archer VR1600v v1, with firmware v2.1.0.

### DIFF
--- a/tplinkrouterc6u/client/mr.py
+++ b/tplinkrouterc6u/client/mr.py
@@ -5,7 +5,7 @@ from urllib.parse import quote
 from requests import Session, Response
 from datetime import timedelta, datetime
 from logging import Logger
-from tplinkrouterc6u.common.helper import get_ip, get_mac, get_value
+from tplinkrouterc6u.common.helper import get_ip, get_ipv6, get_mac, get_value
 from tplinkrouterc6u.common.encryption import EncryptionWrapperMR, EncryptionWrapperMRGCM
 from tplinkrouterc6u.common.package_enum import Connection, VPN
 from tplinkrouterc6u.common.dataclass import (
@@ -15,6 +15,7 @@ from tplinkrouterc6u.common.dataclass import (
     IPv4Reservation,
     IPv4DHCPLease,
     IPv4Status,
+    IPv6Status,
     SMS,
     LTEStatus,
     VPNStatus,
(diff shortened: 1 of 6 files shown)


david9600:
Examples of sensor display, using integration modified to create a wan_ipv6_address sensor. The first image is normal ISP connection (dual stack), the second is a backup service. If IPv6 is either not supported or not active on the connection, a null address is shown.

dual-stack-normal

ipv4-only

tplinkrouterc6u/client/mr.py
```diff
@@ -152,7 +153,7 @@ def get_status(self) -> Status:
         acts = [
             self.ActItem(self.ActItem.GS, 'LAN_IP_INTF', attrs=['X_TP_MACAddress', 'IPInterfaceIPAddress']),
             self.ActItem(self.ActItem.GS, 'WAN_IP_CONN',
-                         attrs=['enable', 'MACAddress', 'externalIPAddress', 'defaultGateway', 'name']),
+                         attrs=['enable', 'MACAddress', 'externalIPAddress', 'X_TP_ExternalIPv6Address', 'defaultGateway', 'name']),
```
AlexandrErohin, line 156: I think it is better not to add X_TP_ExternalIPv6Address to get_status, as older routers may not have this parameter

david9600: If a router doesn't return a certain parameter in response to the WAN_IP_CONN request, it will be reported as a null string.  I tested this by changing the parameter to an unknown name ... but will test again using a different series of router (i.e. different client), and will report results back here in case of interest.

The reason of course for adding the parameter to the get_status method is so it's available to the integration, however, the PR was only a suggestion based on the work I'd already done for my own installation (where I've pointed the integration to a fork of the API containing the changes).

AlexandrErohin: I had an experience with 2 models - when adding extra parameter (for example to WAN_IP_CONN) it breaks the all respond for WAN_IP_CONN. So why I worry to not break get_status for some routers

david9600: Yes, you are right ... I repeated the tests, with HA restarts in between to eliminate caching, and found the problem you mention.
I originally thought it could be fixed in parsing of the response, but when I ran a network trace, it showed the router didn't respond to the remainder of the requested actions, after encountering an unknown parameter.
As you said, it won't be possible to include a request for IPv6 data in the existing Status request.

I will close this PR and rethink the approach.  It will also allow me some time to develop a solution which will work for some other router models.

PR #174, opened 2026-07-24
Add MR600 v2 support for firmware generations using ECIES/P-224 key exchange

drscooto:
Fixes #361.

## Root cause

Archer MR600 has no dedicated client class, so `TplinkRouterProvider` auto-detection was landing on `TplinkC3200Router` — built for a different, older device. Its `supports()`/`authorize()` are loose enough to accept almost any Archer-branded response, and its `req_act()` uses a plaintext Basic-Auth-cookie `/cgi` endpoint with no encryption at all. Firmware `1.10.0` (RC) rejects that plaintext request outright, hence the `IGD_DEV_INFO` failure in #361.

## What changed on this firmware

Verified via live protocol capture against a real MR600 v2 on `1.10.0 0.9.1 v0001.0 Build 260618 RC.40417n` (hooking the router's own JS crypto functions in a browser session, then confirming a from-scratch Python re-implementation against the live device):

- The act/OID request **format is unchanged** — `IGD_DEV_INFO`, `WAN_LTE_LINK_CFG`, etc. all still work exactly as `TPLinkMRClient` already builds them.
- The **crypto changed**: RSA key exchange is replaced with **ECIES over NIST P-224** (the router's own `ecc.min.js` is stock SJCL) for request signing. Session payload is still AES-128-CBC/PKCS7. Each request additionally carries a separate `HMAC-SHA256(payload, key=session_aes_key)` appended to the signature JSON as a 4th field alongside the ECIES `iv`/`ct`/`tag`.
- Login/other responses are plain JSON (`{"ret":0}`) rather than the `$.ret=0;` JS-var format used elsewhere.

## Changes

- `common/encryption.py`: new `EncryptionWrapperMRECC`, mirroring `EncryptionWrapperMR`'s interface but with ECIES-P224 signing instead of RSA, plus the HMAC field.
- `client/mr.py`: new `TPLinkMR600Client(TPLinkMRClient)` — reuses `get_firmware`/`get_status`/`get_lte_status`/SMS/etc. unchanged (they all go through `req_act`/`_fill_acts`, untouched), overriding only the auth/crypto path (`_req_ecc_key`, `authorize`, `_prepare_data`, `_req_login`, `_parse_ret_val`, `logout`).
- `provider.py`: registers `TPLinkMR600Client` in the detection list (ordered before `TplinkC3200Router`, which is what MR600 was incorrectly falling through to).
- `test/test_client_mr600.py`: unit tests for the new encryption wrapper, including round-trip encrypt/decrypt validated against **real captured ciphertext/plaintext pairs** from the live router session, plus `_parse_ret_val` format handling.

## Validation

Live end-to-end test via the exact path a real integration uses:

```python
router = TplinkRouterProvider.get_client(host, password, username)
# -> TPLinkMR600Client (auto-detected)
router.authorize()
router.get_firmware()     # -> real hardware/model/firmware strings
router.get_status()       # -> real LAN/WAN/client data
router.get_lte_status()   # -> real RSRP/RSRQ/SNR/signal data
router.logout()           # -> clean
```

All returned real data from the live device. Full existing test suite (`pytest test`, 319 tests) still passes, plus 7 new tests. `flake8` clean.

## Known limitation

The wrong-password error path (`_req_login`'s `HTTP_ERR_USER_PWD_NOT_CORRECT` handling, which parses `var currAuthTimes=...` from the response) wasn't exercised against this firmware — I didn't want to trigger the router's failed-login lockout counter against a device in daily use to check its exact error format. The happy path is fully verified; that error-formatting edge case may need a follow-up if someone hits it.

I don't have another MR600 to cross-check against, so I can't verify whether this firmware/crypto generation is uniform across all MR600 units or specific to this RC build — happy to iterate if it doesn't match for others.


AlexandrErohin:
@drscooto Thank you for the MR!

tplinkrouterc6u/provider.py
```diff
@@ -74,6 +74,7 @@ def get_clients() -> dict[str, type[AbstractRouter]]:
             TPLinkWR841NClient.__name__: TPLinkWR841NClient,
             TPLinkMRClientGCM.__name__: TPLinkMRClientGCM,
             TPLinkMRClient.__name__: TPLinkMRClient,
+            TPLinkMR600Client.__name__: TPLinkMR600Client,
             TPLinkMR200Client.__name__: TPLinkMR200Client,
             TPLinkMR6400v7Client.__name__: TPLinkMR6400v7Client,
             TPLinkVR400v2Client.__name__: TPLinkVR400v2Client,
```
AlexandrErohin, line 79: Would it work if TPLinkMR600Client add after TPLinkVR400v2Client ?

drscooto:
Good catch — moved it in abe281d, now registered right after `TPLinkVR400v2Client` so the existing MR200/MR6400v7/VR400v2 detection gets first refusal before the new ECC-key probe runs. Full suite still passes (326 tests) and flake8 is clean. Thanks for reviewing!

AlexandrErohin:
@drscooto Thank you!

PR #175, opened 2026-07-25
fix: accept both list shapes for IPv4 reservations and DHCP leases (Archer BE805 v1.20)

LucasIcarus:
### Problem

`get_ipv4_reservations()` raises on an **Archer BE805 v1.20**, firmware **`1.5.1 Build 20260523 rel.11659(5347)`** (resolved client: `TplinkRouterV1_11`):

```
TypeError: string indices must be integers, not 'str'
```

### Cause

That firmware answers `admin/dhcps?form=reservation&operation=load` with an **envelope** rather than a bare array:

```
raw type    : dict
raw keys    : ['list']
raw['list'] : list[21] of dict
element keys: ['comment', 'enable', 'hostname', 'ip', 'key', 'mac']
```

`for item in data:` therefore iterates the dict's **keys**, and the first `item['mac']` subscripts a string. The payload itself is perfectly healthy — all 21 reservations are present with the expected fields.

### Approach — additive, not substitutive

Changing `data` to `data['list']` would fix this model and **break every router that answers with a bare array**. These response shapes are model- and firmware-specific, so the fix accepts either.

A small shared `_as_list()` helper:

- returns the payload unchanged when it is already a list — **no behaviour change for existing devices**
- unwraps `{"<key>": [...]}` when the router wraps it
- otherwise raises `ClientError` naming the keys actually received, which is considerably more useful in a bug report than a `TypeError` three frames deeper

Applied to `get_ipv4_dhcp_leases()` as well. It has the identical pattern and happens to work on this firmware, but leaving the same latent bug next door only defers it, and a bare array passes through untouched.

### Testing

Verified on real hardware: Archer BE805 v1.20, firmware
1.5.1 Build 20260523 rel.11659(5347), client `TplinkRouterV1_11`.
Before: `TypeError: string indices must be integers, not 'str'`.
After: all 21 reservations returned; leases still 15.
323 existing tests pass, 4 added.

Thanks for the library; adding BE805 support made this device usable at all.

### DIFF
--- a/test/test_client_c6u.py
+++ b/test/test_client_c6u.py
@@ -1,4 +1,5 @@
 from unittest import main, TestCase
+from unittest.mock import patch, Mock
 from macaddress import EUI48
 from ipaddress import IPv4Address
 from json import loads
@@ -15,6 +16,7 @@ from tplinkrouterc6u import (
     VpnClientServer,
     VpnClientDevice,
 )
+from tplinkrouterc6u.common.exception import ClientError
 
 
 class TestTPLinkClient(TestCase):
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@LucasIcarus Thank you!

PR #176, opened 2026-07-26
Add get_lte_serving_cells() to TPLinkMRClient

d96moe:
## Summary
- Adds `get_lte_serving_cells()` to `TPLinkMRClient`, returning active LTE/NR
  serving cells (`cellConnectionStatus == '1'`) from `DEV2_LTE_SERVING_CELL_INFO`
- Follows the response-parsing pattern already used elsewhere in `mr.py`
  (`_to_list()` to normalize single-cell dict vs multi-cell list responses)
- Returns raw `list[dict]` rather than a dataclass, since the router's field
  set varies by network type and downstream consumers only need a handful of
  fields per cell type

## Motivation
Follow-up to feedback on AlexandrErohin/home-assistant-tplink-router#367 —
router communication logic (building the ActItem, parsing the response)
belongs in the client library rather than inline in the HA integration's
coordinator.

The integration side can then use this to expose per-cell-type sensors, reading:
- `networkType`, `band`, `ARFCN`, `downBandWidth`, `downFreq`,
  `downlinkModType`, `uplinkModType`, `CQI`, `RI`, `numRbs`
- `RSRP`, `RSRQ` (LTE/LTE+ only)

...separately for the NR cell, the LTE anchor cell, and the LTE+ (carrier
aggregation secondary) cell, plus a derived "active bands" summary
(e.g. `B8+N1`).

## Testing
- Unit tests added for multi-cell and empty-response cases
  (`test/test_client_mr.py`)
- Live-tested against a real NX200 router via a Home Assistant coordinator
  update, confirming NR/LTE band sensors populate correctly

## Test plan
- [x] Unit tests pass
- [x] Live verification against physical hardware

### DIFF
--- a/tplinkrouterc6u/client/mr.py
+++ b/tplinkrouterc6u/client/mr.py
@@ -824,6 +824,11 @@ class TPLinkMRClient(TPLinkMRClientBase):
 
         return status
 
+    def get_lte_serving_cells(self) -> list[dict]:
+        acts = [self.ActItem(self.ActItem.GL, 'DEV2_LTE_SERVING_CELL_INFO')]
+        _, values = self.req_act(acts)
+        return [c for c in self._to_list(values) if c.get('cellConnectionStatus') == '1']
+
 
 # Class for MR series routers which supports AES cipher GCM mode
 class TPLinkMRClientGCM(TPLinkMRClientBaseGCM, TPLinkMRClient):
--- a/tplinkrouterc6u/client/mr.py
+++ b/tplinkrouterc6u/client/mr.py
@@ -18,6 +18,7 @@ from tplinkrouterc6u.common.dataclass import (
     IPv4Status,
     SMS,
     LTEStatus,
+    ServingCell,
     VPNStatus,
 )
 from tplinkrouterc6u.common.exception import ClientException, ClientError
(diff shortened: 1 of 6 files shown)


AlexandrErohin:
@d96moe Thank you for MR!

tplinkrouterc6u/client/mr.py
```diff
@@ -823,6 +823,11 @@ def get_lte_status(self) -> LTEStatus:
 
         return status
 
+    def get_lte_serving_cells(self) -> list[dict]:
```
AlexandrErohin: I think it is better to return list of object then list of dict like get_sms does

d96moe:
Pushed a new commit adding \`get_lte_serving_cells()\` to \`TPLinkEXClient\` as well — previously it only existed on \`TPLinkMRClient\`. Also changed the return type on both to \`list[ServingCell]\` (new dataclass in \`common/dataclass.py\`) instead of \`list[dict]\`, matching the \`get_sms()\` pattern.

Correction to my earlier comment: I claimed this was live-tested via \`TPLinkMRClient\`. That was wrong — I'd edited \`coordinator.py\` but hadn't restarted Home Assistant, so I was actually seeing the *old* code still running, not the new. After a proper restart, the real router turned out to resolve to \`TPLinkEXClient\`, not \`TPLinkMRClient\` — hence adding it there too. \`TPLinkEXClient\` is now the implementation that's actually been verified against real hardware. 

\`TPLinkMRClient\`'s version is kept for consistency (every method here is duplicated across both classes), but it hasn't been tested against a real device, only mock responses in the unit tests. Given that, should I actually pull it out of this PR until someone can verify it against real MR-class hardware, rather than merging unverified code?

tplinkrouterc6u/common/dataclass.py
```diff
@@ -267,6 +264,22 @@ class SMS:
     unread: bool
 
 
+@dataclass
+class ServingCell:
+    network_type: int
```
AlexandrErohin, line 269: can network_type be formated to string like https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/common/dataclass.py#L310 ?

AlexandrErohin:
I had to delete get_lte_serving_cells from TPLinkMRClient as it doesnt have parameters started with DEV2_...

PR #173, opened 2026-07-23
Guest Wi-Fi settings and SG protocol compatibility

mikeluyten:
- Fixed 403 Forbidden error in c6u.py with conditional login signature.
- Guest Wi-Fi management methods and dataclass.
- Updated SG client to use dictionary-based payloads for better form-encoding.

PR #177, opened 2026-07-26
Guest Wi-Fi settings and SG protocol compatibility

mikeluyten:
reimplementation of https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/173, had to re-do to fix local commit history
- Implement generic get_wifi(wifi: Connection) method in TplinkBaseRouter.
- Enhanced set_wifi method with support for granular settings (SSID, hidden, etc).
- Fix 403 Forbidden error in c6u.py via conditional login signature.
- Update SG client to use dictionary-based payloads for better form-encoding.

### DIFF
--- a/test/test_client_c6u.py
+++ b/test/test_client_c6u.py
@@ -1,4 +1,5 @@
 from unittest import main, TestCase
+from unittest.mock import patch, Mock
 from macaddress import EUI48
 from ipaddress import IPv4Address
 from json import loads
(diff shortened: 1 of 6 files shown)


AlexandrErohin:
@mikeluyten it looks like these changes broke integration. https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/369 and https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/363#issuecomment-5113581735

mikeluyten:
Testing a fix now looks like its unhappy when there is no data block (like for the get firmware call)

Issue #171, opened 2026-07-18
Archer AX12 v1.0 (firmware 1.2.3) — authorize() fails with HTTP 403 on form=login


alovir:
Router: Archer AX12 v1.0
Firmware: 1.2.3 Build 20251013 Rel. 44365(4252)
Library version: 5.23.0
Client used: TplinkRouterProvider.get_client() → resolves to TplinkRouter

Description:
authorize() fails consistently with HTTP 403 on the form=login POST,
even though:
- form=keys and form=auth both return 200 with valid data
- Login works fine manually via the browser with the same credentials
- RSA key returned is 1024-bit (256 hex chars), matching the classic
  AES+RSA scheme (not V1_11 / RSA-only)

Debug log:
POST .../login?form=keys&operation=read  -> 200
POST .../login?form=auth&operation=read  -> 200
POST .../login?form=login                -> 403

Already tried:
- Using a shared requests.Session() across keys/auth/login calls
  (in case a sysauth cookie needed to persist) — no change
- Adding an explicit Origin header to the login POST — no change
- Confirmed password is correct (works in browser)
- Confirmed not rate-limited/locked out (browser login succeeds
  immediately after)

I have a browser DevTools capture (Network tab, form=login request/
response headers + payload) available if useful for comparison.

Happy to test a patch if you can point me in the right direction.

mikeluyten:
You can test my varient: https://github.com/mikeluyten/TP-Link-Archer-C6U
I have a pr open for it here https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/173

alovir:
> You can test my varient: https://github.com/mikeluyten/TP-Link-Archer-C6U I have a pr open for it here [#173](https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/173)

Update: tested two more things, both still hit the same 403.

1. Tested @mikeluyten's fork (git+https://github.com/mikeluyten/TP-Link-Archer-C6U.git@main,
   resolves to tplinkrouterc6u 5.24.0). Looked at the diff in PR #173 — the
   conditional signature change (`True if self._logged is False else False`)
   only affects post-login requests, since `_logged` is already `False`
   before the first `authorize()` call, so it evaluates to `True` either
   way at that point — same as before the fix. Confirmed empirically:
   identical 403 on the initial `form=login` POST.

2. Updated router firmware from 1.2.3 Build 20251013 to
[...]

PR #185, opened 2026-08-04
fix: correct OAEP signature chunk size in TplinkRouterSG

Fozar:
## What's broken

`_build_login_signature()` in `TplinkRouterSG` splits the signature string
into fixed 53-byte chunks before RSA-OAEP-encrypting each one. 53 bytes is
sized for PKCS1v1.5's 11-byte overhead — OAEP's overhead
(`2*hLen+2` = 42 bytes with pycryptodome's default SHA-1) is much larger, so
any chunk over `modulus_len-42` bytes overflows `PKCS1_OAEP.encrypt()` with
`ValueError: Plaintext is too long.` on the 512-bit auth key used for the
login signature.

The practical effect: `authorize()` always raises, so `supports()` always
returns `False`, so `TplinkRouterSG` can never actually be selected by
`TplinkRouterProvider.get_client()` or used at all — every SG/CE_RED
certified router falls through to the legacy `TplinkRouter` class, which
uses the wrong hash algorithm and padding scheme for these devices and
fails with a `403`/empty-body response on every login attempt.

## Fix

Compute the OAEP chunk size from the auth key's modulus length instead of
reusing the PKCS1v1.5-sized constant.

## Testing

Verified end-to-end against a real Archer AX12 v1.0 running firmware
certified `SG CLS L1 STAGE2` (previously failing with
`Cannot authorize! Error - Expecting value: line 1 column 1 (char 0)`):

- `TplinkRouterSG(...).supports()` → `True`
- `authorize()` → succeeds, returns a valid `stok`
- `get_firmware()` → real firmware info (`Archer AX12 v1.0`, `1.3.2 Build 20260324...`)
- `get_status()` → real WAN IP
- `logout()` → clean

Related: #171

### DIFF
--- a/test/test_client_sg.py
+++ b/test/test_client_sg.py
@@ -26,6 +26,40 @@ class TestTplinkRouterSGUnit(TestCase):
         client = TplinkRouterSG('http://192.168.0.1', long_password)
         self.assertFalse(client.supports())
 
+    def test_build_login_signature_real_key_size(self) -> None:
+        """Regression test for #171/#185.
+
+        The auth ("nn"/"ee") key TP-Link actually serves is 512-bit (128 hex
+        chars). SIGNATURE_OFFSET (53) is sized for PKCS1v1.5's 11-byte
+        overhead; under OAEP's much larger overhead (2*hLen+2 = 42 bytes for
+        the SHA-1 pycryptodome defaults to) a 53-byte chunk does not fit in a
+        single 512-bit block, and PKCS1_OAEP.encrypt() raises
+        "Plaintext is too long.". Every previous test in this file mocks
+        either `request()` (via TestTPLinkClient) or `_build_login_signature`
+        itself, so this real, unmocked call is the only thing that exercises
+        the actual RSA-OAEP chunking against a realistically-sized key.
+        """
+        client = TplinkRouterSG('http://192.168.0.1', 'testpassword')
+        client._aes_key = '1234567890123456'
+        client._aes_iv = '6543210987654321'
+        client._hash = sha256(b'admintestpassword').hexdigest()
+        client._seq = 555111222
+        # A real 512-bit auth key as served by an Archer AX12 on SG CLS L1
+        # STAGE2 firmware (public key only -- no private counterpart needed
+        # to exercise the encryption path).
+        client._nn = (
+            'ca8f1711cc27576fb0dae0d7df1b6a90465e8ea31ccf46b0004f4c60f6617df'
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@Fozar Thank you for your PR! Im afraid it is not compability with previous router firmware and it breaks integration for them

Fozar:
Thanks for the quick look! Could you share which router model/firmware regressed and what error/behavior you saw? Want to make sure I'm fixing the right thing.

One thing worth noting: this PR only changes `_build_login_signature()` (OAEP RSA chunk size for the login step) — it doesn't touch `_build_request_signature()` or the legacy `TplinkRouter` class, so it shouldn't affect non-SG devices at all.

For SG/CE_RED devices with a larger auth key (e.g. 1024-bit), the old fixed `SIGNATURE_OFFSET=53` chunk size never overflowed OAEP (max payload there is 86 bytes), so it "worked" — just chunked smaller than necessary. My fix computes the OAEP-safe max instead, which is *smaller* than 53 for the 512-bit key on my AX12, but *larger* than 53 for bigger keys — and that size change (more bytes per block, fewer blocks) is the most likely place something could regress if the firmware's signature parsing is sensitive to it.

If that's the failure mode, a safer version would be to cap it: `oaep_step = min(SIGNATURE_OFFSET, rsa_byte_len - 2*20 - 2)`. That preserves the exact old chunking (53 bytes) for any key large enough to support it, and only shrinks the chunk size when 53 would actually overflow (small keys like mine). Happy to push that variant if it helps — just need to know which device hit the regression so I can reason about its key size.

AlexandrErohin:
Unfortunately, I dont have this router serie. The SG class was added here https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/136 by @mattmanning Would be great if someone who uses SG class could test your fix

@mattmanning Could you test this fix with your router please?

Fozar:
Pushed the capped variant: `oaep_step = min(SIGNATURE_OFFSET, rsa_byte_len - 2*20 - 2)`. For any auth key large enough that the old fixed 53-byte chunk never overflowed OAEP (roughly 760-bit+ modulus), this is byte-for-byte identical to the pre-#185 behavior — no change at all for whatever device may have regressed. It only kicks in below that threshold (e.g. my 512-bit AX12 key), where 53 bytes would overflow and the old code always raised `ValueError: Plaintext is too long.` anyway. All 59 existing tests plus the regression test still pass.

@mattmanning if you get a chance to test against your SG-class router, that'd help confirm this doesn't regress anything.

PR #186, opened 2026-08-08
Add IPv6 to get_status in MR client.

david9600:
This adds IPv6 support to get_status in MR client.
A new WAN_IP_CONN request is added at the end of the list of actions, to obtain IPv6 specifics.
By placing it last, it shouldn't affect responses earlier in the list, for routers without IPv6 support.  This has been tested in principle by specifying an invalid parameter name - there was no impact on responses to actions earlier in the list, and the router continued to respond to data fetches.

Testing was done with VR1600v v1, f/w 2.1.0, and VR400 v3, f/w 1.6.0.

Requests and responses for the test cases:
[VR400 v3 - test invalid parm.txt](https://github.com/user-attachments/files/30852204/VR400.v3.-.test.invalid.parm.txt)
[VR400 v3 - IPv6 disabled.txt](https://github.com/user-attachments/files/30852209/VR400.v3.-.IPv6.disabled.txt)
[VR400 v3 - IPv6 enabled.txt](https://github.com/user-attachments/files/30852214/VR400.v3.-.IPv6.enabled.txt)
[VR1600v - IPv6 enabled.txt](https://github.com/user-attachments/files/30852218/VR1600v.-.IPv6.enabled.txt)

### DIFF
--- a/tplinkrouterc6u/client/mr.py
+++ b/tplinkrouterc6u/client/mr.py
@@ -169,6 +169,8 @@ class TPLinkMRClientBase(AbstractRouter):
                 'X_TP_TotalPacketsSent',
                 'X_TP_TotalPacketsReceived',
             ]),
+            self.ActItem(self.ActItem.GS, 'WAN_IP_CONN',
+                         attrs=['enable', 'X_TP_IPv6Enabled', 'X_TP_IPv6ConnStatus', 'X_TP_ExternalIPv6Address']),
         ]
         _, values = self.req_act(acts)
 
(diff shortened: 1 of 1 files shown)


david9600:
If adding the request to the tail of the existing list of actions is a concern though, I can do it another way - especially as I've found that the existing WAN code (for IPv4 & IPv6) doesn't fully support a router with WAN failover, i.e. dual WAN links.

During testing, I found there can be circumstances where both WAN interfaces return 'enable=1'.  The existing code treats them the same, so the for-loop returns the last one, which is usually not the primary service, but the USB backup modem.  I'm intending to rewrite the WAN code block so it can distinguish between primary and backup services, in routers having that capability.

tplinkrouterc6u/client/mr.py
```diff
@@ -169,6 +169,8 @@ def get_status(self) -> Status:
                 'X_TP_TotalPacketsSent',
                 'X_TP_TotalPacketsReceived',
             ]),
+            self.ActItem(self.ActItem.GS, 'WAN_IP_CONN',
```
AlexandrErohin, line 172: Im afraid, It could break compability with other routesr. MR class uses a lot of different routers. When I was developing this class I got a problem. If I add a field (for example X_TP_IPv10Enabled) that doesnt have router - I got one single error in the response. So adding WAN_IP_CONN to the main request may lead to same result. I would suggest to do something like that https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/c6u.py#L431 add another request to try catch to avoid braking

david9600:
Closing this PR to try a different approach.

PR #187, opened 2026-08-08
Add EasyMesh requests to c6u client.

david9600:
This adds EasyMesh requests to get_status in c6u client, and provides AP association information for wifi devices.

For each wifi device, the AP name is stored as an extra attribute in the device tracker object.
To simplify sorting in a dashboard display, I've prefixed the main node name with an asterisk, so the devices attached to that node can be listed as a group, before the groups of devices on each satellite node. This was a simple way of differentiating between main and satellite without adding another attribute field.

It was found during testing with AX55 that the existing 'signal' value for devices connected to a satellite node is sometimes the value at the satellite node, but at other times it appears to be left over from when it was on the main node.  The code I've written places the value provided by the satellite node into the existing 'signal' attribute.

Below is a small piece of the HA dashboard display, created with jinja templating to select and sort attributes.
The columns are device name, ap name, signal level, and last octet of ip address.

easymesh-dashboard-clip

Later, if the changes are acceptable, I will submit a PR on the main integration for an extra device attribute to contain the ap_name.

### DIFF
--- a/tplinkrouterc6u/client/c6u.py
+++ b/tplinkrouterc6u/client/c6u.py
@@ -471,6 +471,22 @@ class TplinkBaseRouter(AbstractRouter, TplinkRequest):
             # WiFi might be disabled on the router, skip wireless statistics
             pass
 
+        try:
+            easymesh_node_list = self.request('admin/easymesh_network?form=get_mesh_device_list_all&operation=read', 'operation=read')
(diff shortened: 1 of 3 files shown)


david9600:
I've just realised that I've based this branch on my repo which already contained changes for another PR.
I will close this PR and open another, using the master C6U as the base.

PR #188, opened 2026-08-08
Add EasyMesh requests to c6u client.

david9600:
This adds EasyMesh requests to get_status in c6u client, giving AP association information for wifi devices.

For each wifi device, the AP name is stored as an extra attribute in the device tracker object.
To simplify sorting in a dashboard display, I've prefixed the main node name with an asterisk, so the devices attached to that node can be listed as a group, before the groups of devices on each satellite node. This was a simple way of differentiating between main and satellite without adding another attribute field.

It was found during testing with AX55 that the existing 'signal' value for devices connected to a satellite node is sometimes the value at the satellite node, but at other times it appears to be left over from when it was on the main node. The code I've written places the current value provided by the satellite node into the existing 'signal' attribute.

Below is an example of a dashboard display which can be created with jinja templating from the device attributes.
The columns are device name, ap name, signal level, and last octet of ip address.

easymesh-dashboard-clip

Later, if the changes are acceptable, I will submit a PR on the main integration for an extra device attribute to contain the ap_name.

### DIFF
--- a/tplinkrouterc6u/client/c6u.py
+++ b/tplinkrouterc6u/client/c6u.py
@@ -471,6 +471,22 @@ class TplinkBaseRouter(AbstractRouter, TplinkRequest):
             # WiFi might be disabled on the router, skip wireless statistics
             pass
 
+        try:
+            easymesh_node_list = self.request('admin/easymesh_network?form=get_mesh_device_list_all&operation=read', 'operation=read')
+            for ap in easymesh_node_list:
+                # 'sclient' is mesh main or satellite, 'nclient' is a network device
+                sclient_detail = self.request('admin/easymesh_network?form=mesh_sclient_detail&operation=read&mac='+ap['mac'], 'operation=read&mac='+ap['mac'])
+                for nclient in sclient_detail['mesh_nclient_list']:
+                    if ap['role'] == 'satellite_router':
+                        devices[nclient['mac']].ap_name = ap['name']
+                        devices[nclient['mac']].signal = nclient['signal_strength']
+                    else:
+                        # prefix * helps sorting main node devices in display
+                        devices[nclient['mac']].ap_name = '*'+ap['name']
+        except Exception:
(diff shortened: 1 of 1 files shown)


david9600:
These are the AX55 responses (edited to remove device idents).
[EasyMesh-device-list-responses.txt](https://github.com/user-attachments/files/30865481/EasyMesh-device-list-responses.txt)
[EasyMesh-sclient-detail-responses.txt](https://github.com/user-attachments/files/30865483/EasyMesh-sclient-detail-responses.txt)

tplinkrouterc6u/client/c6u.py
```diff
@@ -471,6 +471,22 @@ def _add_device(conn: Connection, item: dict) -> None:
             # WiFi might be disabled on the router, skip wireless statistics
             pass
 
+        try:
```
AlexandrErohin: I would add something like https://github.com/david9600/TP-Link-Archer-C6U/blob/682cb2d281285ae53fd374e76f92f86189f341b6/tplinkrouterc6u/client/c6u.py#L436

add block IF - so if router doesnt have this route the next run would not call this route again

david9600: Thank you for reviewing my code, and for the very sensible recommendation!
I have updated the logic.

AlexandrErohin:
@david9600 Thank you! Could you test my little edditions for your code in main branch please?

david9600:
I started with a clean slate by deleting the integration entry, redownloading, and restarting HA.
Then updated the manifest requirements to "tplinkrouterc6u@git+https://github.com/AlexandrErohin/TP-Link-Archer-C6U.git@main", and added the 'ap_name' extra attribute in device_tracker.py.

Then restarted HA, and added the integration (for AX55 EasyMesh main node).
The EasyMesh display appears to be working normally after the resilience improvements!

AlexandrErohin:
@david9600 Thank you!

PR #189, opened 2026-08-11
Fix signal/missing-IP handling and expose ap_name, ssid, frequency for R series

rz467fzs7d:
## Summary

Adds AP/SSID/frequency data to `Device` for TP-Link R series (AC controller) clients, and fixes two bugs in `TPLinkRClient.get_status()` that affect real devices.

### Bug fixes

1. **Signal strength was never populated.** The guard checks for `signal` but reads `item['rssi']`:
   ```python
   if 'signal' in item:
       dev.signal = int(item['rssi'])
   ```
   The R series `host_info` table returns `rssi`, not `signal`, so the branch never runs. Now checks both keys.

2. **Clients without an IP raised an exception.** `get_ip(item['ip'])` assumed `ip` is always present. Offline or newly-associated clients can omit it, which aborts the entire poll. `Device._ipaddr` is now `IPv4Address | None` and `ipaddr` returns `None` accordingly.

### New fields

`Device` gains three optional fields, populated when the router reports them: `ap_name`, `ssid`, `frequency`. `freq_name` was already parsed to derive `Connection` type but discarded; it is now retained. These matter for multi-AP setups where knowing which AP a client is associated with is the useful signal.

All additions are optional with `None` defaults, so existing consumers are unaffected.

## Test plan

- Extended `test/test_client_r.py` to assert `ap_name`, `ssid`, `frequency`, and `signal` on a wireless client
- Removed `ip` from one fixture entry to cover the missing-IP path, asserting `ipaddr`/`ipaddress` are `None`
- Full suite passes: 314 tests, no failures

Verified against a live TP-Link AC controller, where both bugs reproduce on the current `main`.

### DIFF
--- a/test/test_client_r.py
+++ b/test/test_client_r.py
@@ -331,7 +331,6 @@ $(document).ready(function(e){
           ".name": "2EDE61E7EA70",
           "up_speed": "0",
           "host_save": "on",
-          "ip": "192.168.60.212",
           "time_obj": "any",
           "is_cur_host": "false",
           "limit": "0",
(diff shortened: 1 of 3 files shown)


AlexandrErohin:
@rz467fzs7d Hi! Could you update your Pull Request with main branch as some of your eddings has been alrerady merged here https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/166/changes

PR #190, opened 2026-08-11
fix(sg): VPN client status broken on BE-series routers (missing header + WireGuard)

reformy:
## Problem

On my Archer BE3600 (`TplinkRouterSG` client), `get_vpn_client_status()` always
failed, so the VPN client entities never showed up.

Two separate bugs caused this:

1. **Missing `Content-Type` header on read requests.** `request()` only set
   `Content-Type: application/x-www-form-urlencoded` for write operations.
   The BE-series firmware requires it on every request, so read-only calls
   like `admin/vpn?form=enable` / `form=server` / `form=vpn_user_list` were
   rejected by the router's Lua dispatcher.
2. **WireGuard wasn't a recognized VPN protocol.** `VpnClientServerProtocol`
   only had `openvpn`/`pptp`/`l2tp`, but current firmware offers WireGuard
   in the VPN Client menu. Since the parser raises on any unrecognized
   `type`, a single WireGuard profile broke the whole status call.

## Fix

- Set the `Content-Type` header on the read branch of `request()` too.
- Add `WIREGUARD = 'wireguard'` to `VpnClientServerProtocol`.

Added regression tests in `test/test_client_sg.py` for both: one asserting
the header is present on reads, one feeding a WireGuard server dict through
`get_vpn_client_status()`.

Two independent fixes, kept as two commits for easier review.

### DIFF
--- a/test/test_client_sg.py
+++ b/test/test_client_sg.py
@@ -3,7 +3,12 @@ from hashlib import sha256
 from unittest import main, TestCase
 from unittest.mock import patch, Mock
 
-from tplinkrouterc6u import TplinkRouterSG, ClientException
+from tplinkrouterc6u import (
+    TplinkRouterSG,
+    ClientException,
+    VpnClientServer,
+    VpnClientServerProtocol,
+)
 from test_client_c6u import TestTPLinkClient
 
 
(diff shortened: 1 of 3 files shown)


AlexandrErohin:
@reformy Thank you!

reformy:
> @reformy Thank you!

Oh no - thank you for this project.

AlexandrErohin:
Would be great if you could give a star to the repo :) Thanks

PR #193, opened 2026-08-12
fix: add TplinkRouterAX72 for firmware requiring RSA-OAEP login signatures

gvneshv:
## Problem
The Archer AX72's firmware signs login requests with nested RSA-OAEP chunking instead of the flat PKCS1v1.5 chunking TplinkRouter/TplinkEncryption assumes.
The existing chunk size (53 bytes, sized for PKCS1v1.5's 11-byte overhead) is far larger than what fits under OAEP's overhead (2*hLen+2 = 42 bytes) on this device's 512-bit auth key, so every login attempt gets a 403.

## Fix
Adds TplinkRouterAX72(TplinkRouter), overriding only the three pieces that actually differ: SHA256 (not MD5) password hashing, RSA-OAEP (not PKCS1v1.5) login-signature encryption, and no trailing `&confirm=true`.
authorize() itself, its 403-retry logic, and every read endpoint are inherited unchanged. supports() does a cheap auth-key-size pre-filter (matching TplinkRouterV1_11's pattern) before attempting a real login.

Confirmed end-to-end (login, get_ipv4_status, get_status, get_firmware, logout) against a real Archer AX72.

## On the chunking shape (re: #185)
This isn't a straight port of TplinkRouterSG's fix.
SG's fix chunks the signature string directly into OAEP-sized pieces. The AX72, as confirmed by capturing and replaying a real login, expects nested chunking instead - split into 53-byte outer pieces first, each further split into OAEP-safe inner pieces.
These produce different block boundaries past the first 53 bytes,
and only the nested form has been confirmed against real hardware, so I kept it rather than simplifying to match SG's shape unverified.

## Testing
- New test/test_client_ax72.py: inherits the shared get_firmware/get_status/ etc. tests from TestTPLinkClient, plus unit tests for supports(), _get_login_data(), _prepare_data()'s SHA256 hashing, and a test_build_login_signature_real_key_size regression test mirroring #185's (reuses the same public 512-bit key already in test_client_sg.py, since key size is what the math depends on, not the specific device).
- Full suite (433 tests) passes locally; flake8 clean.

### DIFF
--- a/test/test_client_ax72.py
+++ b/test/test_client_ax72.py
@@ -1,0 +1,122 @@
+from unittest import main, TestCase
+from unittest.mock import patch, Mock
+
+from tplinkrouterc6u import TplinkRouterAX72
+from test_client_c6u import TestTPLinkClient
+
+
+class TestTPLinkClientAX72(TestTPLinkClient):
+    """Inherits get_firmware/get_status/get_ipv4_status/etc.
+    All other tests unchanged from TestTPLinkClient.
+
+    TplinkRouterAX72 extends TplinkRouter directly (unlike TplinkRouterSG, which extends TplinkBaseRouter),
+    so it inherits TplinkRouter's shortened URL scheme too - no path overrides needed here, unlike TestTPLinkClientSG's firmware_path/game_accelerator_path/etc.
+    """
+
+    router_class = TplinkRouterAX72
+
+
+class TestTplinkRouterAX72Unit(TestCase):
+    """Unit tests specific to TplinkRouterAX72's login signature and detection logic.
+
+    authorize() itself, its 403-retry logic, and password encryption are all inherited unchanged from TplinkEncryption/TplinkRouter
+    and are already covered by the base class's own tests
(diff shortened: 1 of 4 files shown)


AlexandrErohin:
@gvneshv Thank you! I have merged your PR and I have made some edditions for your code. Could you finnaly test TplinkRouterProvider.get_client(), authorize(), get_status(), from main branch ?

gvneshv:
@AlexandrErohin 
Confirmed on my Archer AX72 against current main: TplinkRouterProvider.get_client() -> authorize() -> get_status() -> logout() all succeed. One thing worth flagging: get_client() actually returns TplinkRouterSG for my device (not TplinkRouterAX72) - my router apparently carries SG/CE_RED certification and TplinkRouterSG handles it fully end-to-end. TplinkRouterAX72 was genuinely necessary when I built it against an older tplinkrouterc6u version, but it looks like current main's SG detection may already cover devices like mine. Not sure if that changes anything on your end, just wanted you to have the full picture.

AlexandrErohin:
@gvneshv Thank you for testing! Wow I thought TplinkRouterSG is not supported for your router

PR #195, opened 2026-08-13
Add IPv6 to get_status for MR client.

david9600:
This adds support for IPv6 to MR client get_status, using an additional WAN_IP_CONN request to the router.
If the router does not provide an expected response, no more IPv6 requests are sent.
Tested working with VR1600v v1, f/w 2.1.0, and VR400 v3, f/w 1.6.0.

[IPv6-responses-mr-style.txt](https://github.com/user-attachments/files/31110462/IPv6-responses-mr-style.txt)


AlexandrErohin:
@david9600 Thank you!

PR #196, opened 2026-08-14
Fix OAEP chunk size for small RSA keys

leandrogao86:
What does this PR do?

Fixes RSA-OAEP signature chunking for routers using small RSA keys.

With OAEP-SHA1, the maximum plaintext size is limited by the RSA modulus size. Using the fixed SIGNATURE_OFFSET value can therefore cause Plaintext is too long errors with smaller RSA keys.

This change calculates the maximum OAEP-safe chunk size from the RSA modulus and uses the smaller of that value and the existing SIGNATURE_OFFSET.

For sufficiently large RSA keys, the existing 53-byte chunk size is preserved.

Why?

I encountered this issue while adding Archer AX55 VPN client support. The router uses the SG/CE-RED encryption scheme, and login failed because the OAEP signature exceeded the maximum plaintext size for the RSA key.

This change fixes the issue while preserving the existing behavior for routers with larger RSA keys.

### DIFF
--- a/tplinkrouterc6u/client/sg.py
+++ b/tplinkrouterc6u/client/sg.py
@@ -164,8 +164,16 @@ class TplinkRouterSG(TplinkBaseRouter):
         sign_str = '{}&h={}&s={}'.format(
             self._get_aes_formatted_key(), self._hash, self._seq + data_len)
         sign = ''
-        for i in range(0, len(sign_str), SIGNATURE_OFFSET):
-            chunk = sign_str[i:i + SIGNATURE_OFFSET]
+
+        # OAEP-SHA1 has a maximum plaintext size of
+        # modulus_bytes - 2 * hash_len - 2. Keep the legacy
+        # SIGNATURE_OFFSET for larger keys while avoiding
+        # "Plaintext is too long" errors with smaller RSA keys.
+        rsa_byte_len = len(self._nn) // 2
+        oaep_step = min(SIGNATURE_OFFSET, rsa_byte_len - 2 * 20 - 2)
+
+        for i in range(0, len(sign_str), oaep_step):
+            chunk = sign_str[i:i + oaep_step]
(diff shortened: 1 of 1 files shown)


AlexandrErohin:
@leandrogao86 Thank you for your Pull Request! It looks like it is already added into main branch by this Pull Request https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/185/changes - Could you test please v5.30.0 of this client?

leandrogao86:
Hi Alexandr, I tested v5.30.0 against my Archer AX55.
Authentication now works correctly, so PR #185 indeed fixes the OAEP issue.
However, get_vpn_client_status() still fails on my AX55 with:
Request admin/vpn?form=enable - Response {"data":""}.

So the OAEP/login issue is fixed in v5.30.0, but the VPN Client endpoint still doesn't seem to work with my AX55 firmware.

AlexandrErohin:
As I dont have this router model - Would be great if you could create Pull Request with fix for it

leandrogao86:
Hi Aleksandr, 

I tested an Archer AX55 running firmware 5.30.0 with the current TplinkRouterSG implementation.

The router is correctly detected as TplinkRouterSG, and the VPN Client API works with WireGuard profiles. I was able to:

retrieve the VPN Client status;
enumerate all WireGuard server profiles;
enable/disable the VPN Client;
switch the active WireGuard profile;
and verify that a Surfshark WireGuard connection reaches connected.

So it looks like AX55 firmware 5.30.0 is already supported by the current SG implementation, including WireGuard.

One separate issue I noticed: after TplinkRouterProvider.get_client() returns the TplinkRouterSG instance, calling get_vpn_client_status() immediately raises Not authorised. Calling router.authorize() first makes it work. This seems related to the provider's supports()/authentication lifecycle rather than WireGuard itself.

I haven't opened a PR because I don't think an additional WireGuard implementation is necessary at this point.

PR #197, opened 2026-08-15
Add IPv6 to get_status (with & without wifi) for c80 client.

david9600:
Tested with Archer C24 2.0 f/w 1.12.20 (including tests with invalid request number to simulate no IPv6 support).

[IPv6-responses-c80-style.txt](https://github.com/user-attachments/files/31110446/IPv6-responses-c80-style.txt)


AlexandrErohin:
@david9600 Thank you!

PR #198, opened 2026-08-15
fix: detect TL-SG108E by hardware and populate LAN MAC

prvashisht:
## Problem

On a TL-SG108E v6.0, auto-detection can fail when the Device Description is customized because `supports()` checks `g_title` and `descriStr`, but not the hardware-reported `hardwareStr`.

When detection does succeed, `get_status()` leaves `lan_macaddr` unset. This breaks consumers such as the Home Assistant integration, which uses it as the config entry unique ID.

## Fix

* Check `hardwareStr` for the existing `TL-SG` identifier before falling back to `descriStr`.
* Populate `Status._lan_macaddr` from the switch's `macStr` on a best-effort basis.
* Add regression tests for both cases.

The existing `g_title` and `descriStr` detection paths are unchanged.

## Testing

Tested on:

* TL-SG108E v6.0

Verified on the physical switch with Device Description `TPLink-Main-Switch`:

* `TplinkRouterProvider.get_client()` detects `TPLinkSG108EClient`
* authorization and status retrieval succeed
* `status.lan_macaddr` contains the real switch MAC
* Home Assistant config flow completes successfully
* integration remains healthy after a normal Home Assistant Core restart

Also verified:

* `pytest test/test_client_sg108e.py` — 9 passed
* `pytest test` — 447 passed
* flake8 clean

### DIFF
--- a/test/test_client_sg108e.py
+++ b/test/test_client_sg108e.py
@@ -69,6 +69,31 @@ class TestTPLinkSG108EClient(TestCase):
 
         self.assertTrue(client.supports())
 
+    def test_supports_by_hardware_str_with_custom_description(self) -> None:
+        # Device Description is user-editable, hardwareStr is not.
+        client = TPLinkSG108EClient('http://192.0.2.23', 'password')
+        client._session = Mock()
+        client._session.cookies = Mock()
+        client._session.cookies.clear = Mock()
+
+        root = Mock()
+        root.status_code = 200
+
+        sysinfo = Mock()
+        sysinfo.status_code = 200
+        sysinfo.text = """<html><script>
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@prvashisht Thank you! I have added caching mac adress. Could you test it please?

prvashisht:
Tested current `main` at `f57953a` on my TL-SG108E v6.0, firmware `1.0.0 Build 20230218 Rel.50633`.

`TplinkRouterProvider.get_client()` correctly returns `TPLinkSG108EClient`.

I ran three consecutive `get_status()` calls: all returned the same valid LAN MAC, 8 ports and 5 current links up. Instrumenting `device_info()` showed it was called exactly once across all three status calls, with `_lan_mac_resolved=True` and the MAC cached correctly.

I also repeated the test using `TPLinkSG108EClient` directly with the same result.

Looks good — caching works as intended.

Issue #159, opened 2026-05-19
REQUEST: Toggle DHCP Service


MontrealPaul:
Request: a "set" function, to enable or disable the DHCP service.

Rationale: The set_ipv4_status() function already exposes lan_ipv4_dhcp_enable (bool) and get_ipv4_dhcp_leases, so I was thinking something like set_ipv4_dhcp(bool), or simply set_dhcp(bool).

My platform: I have a TP-Link ArcherC5400X v1.0, firmware 1.2.1 Build 20250928 rel.12879(5553).

Reason: Because my router's DHCP interface and 'management' is not good (terrible, in fact), and is limited to 30 reservations, I instead use my Pi-Hole's DHCP server (combined with another 'add-on' I wrote in Python which makes management easier). Sometimes, however, I must do other things on that server, which make its DHCP service unavailable, so I would like to toggle DHCP on and off.

Thank you!

AlexandrErohin:
It has been added. please try the latest version

MontrealPaul:
Thank you, @AlexandrErohin and @david9600! 
Unfortunately, it doesn't work for me on my TP-Link ArcherC5400X.

set_ipv4_dhcps() raises ClientError: no such callback on the C5400X v1.0, firmware 1.2.1 Build 20250928. The request being sent is admin/dhcps?form=setting but the router returns {'errorcode': 'no such callback', 'success': False}. The C5400X apparently uses a different endpoint.
My script uses **`tplinkrouterc6u import (TplinkC5400XRouter, etc...)`**
N.B.: FWIW, lan_ipv4_dhcp_enable _does_ work. I toggled DHCP on and off in the GUI, and each time lan_ipv4_dhcp_enable  correctly reported the status.

david9600:
Thanks Paul for the feedback. Does the router let you view the data with browser developer tools, when enabling/disabling the DHCP server through the web management interface?  Some do, but some use encryption within the browser, making it a little more difficult.  If you can see the request, then it may be possible to alter the code to accommodate whatever difference there is between that and the AX55 which was used during testing.

MontrealPaul:
I'm not sure if I'm doing this right, but the POST is:
https://tplinkwifi.net/cgi-bin/luci/;stok=[REDACTED]/admin/dhcps?form=setting
Under Request, the "raw" code is:
[...]

PR #199, opened 2026-08-16
Add set_ipv4_dhcps (DHCP server toggle), c6u client.

david9600:
This allows DHCP server function to be enabled/disabled for the LAN (IPv4) in c6u client.
It may answer issue #159.

Tested working with AX55 f/w 1.5.12.
If the change is accepted, I will raise a PR to add corresponding switch in the main integration.

[dhcps form=setting.txt](https://github.com/user-attachments/files/31110898/dhcps.form.setting.txt)


AlexandrErohin:
@david9600 Thank you! I had to rename the method - so it looks like same with parameter lan_ipv4_dhcp_enable from getStatus

Issue #201, opened 2026-08-16
Archer AX73 V2.0 (fw 1.2.1-series): SG client OK for status, but get_firmware/get_ipv4_dhcp_leases/get_ipv4_reservations/get_ipv6_status fail + get_wifi ValueError on host bands


risckyaw:
### Device
- **Model:** Archer AX73 V2.0
- **Firmware:** 1.2.1-series Build 2025+ (login page shows only "Local Password" field, no username — same firmware family as #220 where an AX73 v2.0 with `1.2.1 Build 20250717 rel.40826(4555)` was reported). Exact build not readable via API since `get_firmware()` fails (see below).
- **Library:** `tplinkrouterc6u` 5.30.0 (pip), plain HTTP over LAN
- **Client picked:** `TplinkRouterSG` (via `TplinkRouterProvider.get_client`)

### Working ✅
- `authorize()` / `logout()`
- `get_status()` — full status incl. devices, wifi states, cpu/mem, wan/lan ip
- `get_ipv4_status()` — wan/lan ip, dns, netmask
- `get_wifi()` for `GUEST_2G` / `GUEST_5G` (ssid, psk, encryption) and `IOT_*` (empty config)
- `get_vpn_client_status()`

### Failing ❌ — `ClientError: TplinkRouterSG - Response with error` (encrypted error payload)
All raise the same error shape: `Response {"data":"<base64 blob>"}`:

- `get_firmware()` → `admin/firmware?form=upgrade&operation=read`
- `get_ipv4_dhcp_leases()` → `admin/dhcps?form=client&operation=load`
- `get_ipv4_reservations()` → `admin/dhcps?form=reservation&operation=load`
- `get_ipv6_status()` → `admin/network?form=wan_ipv6_dynamic&operation=read`

### Bug: `get_wifi()` on host bands
`get_wifi(Connection.HOST_2G)` and `get_wifi(Connection.HOST_5G)` raise:

```
ValueError: invalid literal for int() with base 10: 'auto'
```

This firmware returns `channel: 'auto'` for the host bands (guest bands return `channel: null` and parse fine), so the int-cast in the WifiStatus parser breaks.

### Notes
- Happy to run any test script / capture logs (I have the device on hand).
- Context: the HA integration (`home-assistant-tplink-router`) also hits the firmware endpoint on every poll, so the `get_firmware()` failure means Model/Firmware sensors show unknown on this device.

risckyaw:
[...]

Issue #200, opened 2026-08-16
Header addition breaks reads on many modems


mikeluyten:
I believe the addition of the header on the following commit breaks most reads on many modems [764a75d](https://github.com/AlexandrErohin/TP-Link-Archer-C6U/commit/764a75d37c931b7bc9da25d332513a2bd4b3f4fe#r196416683) 

https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/379
https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/380
https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/381

AlexandrErohin:
Please try the latest version

mikeluyten:
Fixed main issue, IOT wifi total clients is now unavailable

AlexandrErohin:
It looks like your router doesnt retun this info. Would be great if you could test get_status method from https://github.com/AlexandrErohin/TP-Link-Archer-C6 and create Pull Request with the fix

mikeluyten:
> It looks like your router doesnt retun this info. Would be great if you could test get_status method from https://github.com/AlexandrErohin/TP-Link-Archer-C6 and create Pull Request with the fix

Will get the status later today, however it worked in the previous version.

mikeluyten:
> > It looks like your router doesnt retun this info. Would be great if you could test get_status method from https://github.com/AlexandrErohin/TP-Link-Archer-C6 and create Pull Request with the fix
> 
> Will get the status later today, however it worked in the previous version.

Looks like its returning it in the get_status, and after restarting home assistant, suddenly its working again... strange, ALL other senors were displaying fine except that one Total IOT... thanks for looking Alexandr!

PR #194, opened 2026-08-13
fix(sg): drop body params duplicated in the path query string

reformy:
Fixes integration not loading in HomeAssistant.

PR #202, opened 2026-08-18
fix(c6u): handle 'auto' wifi channel without crashing

dennismdejong:

Addresses #201 (the wifi-channel part).

You are right that #200 is fixed (by #194) — I have dropped the Content-Type change from this PR since it is not needed. This PR now contains only the `channel: auto` fix:

Firmware on host bands returns `channel: "auto"` for auto-selected channels. `int("auto")` raised `ValueError`, breaking `get_wifi(HOST_2G)` / `get_wifi(HOST_5G)` on the AX73. Non-numeric channels now map to `None`.

## Tests

- `test_get_wifi_channel_auto` — `"auto"` channel yields `channel=None`

Full suite passes, flake8 clean.

### DIFF
--- a/test/test_client_c6u.py
+++ b/test/test_client_c6u.py
@@ -1908,6 +1908,22 @@ class TestWifiGeneric(TestCase):
         self.assertEqual(info.ssid, 'Host_2G')
 
     @patch('tplinkrouterc6u.client.c6u.post')
+    def test_get_wifi_channel_auto(self, mock_post):
+        # Firmware returns channel 'auto' for auto-selected channels; must map to None
+        mock_data = {
+            'wireless_2g_enable': 'on',
+            'wireless_2g_ssid': 'Host_2G',
+            'wireless_2g_channel': 'auto'
+        }
+
+        with patch.object(self.client, 'request', return_value=mock_data):
+            info = self.client.get_wifi(Connection.HOST_2G)
+
+        self.assertTrue(info.enable)
+        self.assertEqual(info.ssid, 'Host_2G')
+        self.assertIsNone(info.channel)
+
+    @patch('tplinkrouterc6u.client.c6u.post')
     def test_set_wifi_enhanced(self, mock_post):
         # Verify set_wifi still builds correct data strings
         with patch.object(self.client, 'request') as mock_request:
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@dennismdejong Error with get_firmware(), get_ipv4_dhcp_leases(), get_ipv4_reservations() and get_ipv6_status() was fixed here https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/194. Is this a new error? Have you test the latest version?

dennismdejong:

Thanks for the review. I checked #194 and it is indeed already merged and present in this PR branch — it is not being regressed by this change. I believe this is a **second, distinct** problem rather than the same one #194 fixed.

**What #194 fixed:** the router rejects a form param that appears in both the path query string and the body (duplicate-key error), breaking reads on `admin/firmware?form=upgrade&operation=read` etc. That was the original integration-not-loading failure.

**What this PR addresses:** the `Content-Type` header added in #190 (commit 764a75d) is now sent on **every** SG read. On several SG-certified routers this makes the firmware/DHCP/reservation/IPv6 endpoints return a response whose decrypted body has no `success: true`, so `_is_valid_response()` returns `False` and `sg.py:306` raises `ClientError: Response with error`.

That distinct failure is still reported open after #194:
- home-assistant-tplink-router #379 (still OPEN) — `get_firmware()` fails at `sg.py:306` on Archer AX55 & BE550, regression between 2.32 (OK) and 2.34 (KO)
- #380 (BE3600) — same `sg.py:306` `Response with error`, payload decrypts fine but lacks the `success` flag
- issue #200 (OPEN) attributes the breakage specifically to the header commit

Those tracebacks are `Response with error` at `sg.py:306` (this PR fixes), not the duplicate-param error (#194 already fixed).

Can you confirm whether you can reproduce `get_firmware()` failing with the latest version on an SG/BE-series router? If so, would you accept this PR (scoping Content-Type to `admin/vpn?` reads only)?

AlexandrErohin:
https://github.com/AlexandrErohin/TP-Link-Archer-C6U/issues/200 confirmed that problem is fixed. So I think the problem with read is fixed. Maybe I miss something?

dennismdejong:

You are right — #200 is confirmed fixed (by #194), so the Content-Type change in this PR was not needed. I have dropped it entirely; this PR now contains only the `channel: "auto"` fix from #201 (which is still open and not addressed by #194).

- **sg.py**: unchanged from main.
- **c6u.py**: `get_wifi()` no longer crashes when firmware returns `channel: "auto"` — maps to `None`.

Would you accept this narrower PR?

AlexandrErohin:
@dennismdejong Thank you!

Issue #158, opened 2026-05-16
set_wifi() crashes on Deco


gil906:
# `set_wifi()` always fails on Deco BE25 (firmware 1.1.7)

Hi, thanks for this library, it's been the only working Python client I've found for Deco.

I hit a wall using `set_wifi()` against a pair of Deco BE25s. Short version: looks like a firmware bug on the router, not on the library side, but the library swallows the real error so it's hard to figure out.

## Setup

- `tplinkrouterc6u==5.19.0`, Python 3.11
- 2x Deco BE25 (hw v1.0), mesh of master + 1 slave
- Both on firmware `1.1.7 Build 20260114 Rel. 14971`
- Tested directly against master and slave, same failure on both.

## Repro

```python
c = TPLinkDecoClient(host="http://<router>", password="<pw>", username="admin")
c.authorize()
c.set_wifi(Connection.GUEST_2G, True)
```

Same call works fine from the Deco mobile app. From Python it raises `An unknown response - Expecting value: line 1 column 1 (char 0)`. Reads, login and reboot all work. Only writes against `admin/wireless?form=wlan` blow up.

The 500 body is still AES-encrypted (no JSON envelope, just raw ciphertext), so `response.json()` fails and the real error gets eaten. Decrypting it manually:

```
/usr/lib/lua/luci/dispatcher.lua:498: Failed to execute call dispatcher target for entry '/admin/wireless'.
The called action terminated with an exception:
?:0: attempt to index global 'mode' (a nil value)
stack traceback:
    [C]: in function 'assert'
    /usr/lib/lua/luci/dispatcher.lua:498: in function 'dispatch'
```

## Why I think it's firmware-side

The keyword in that trace is **global**. Other handlers on the same firmware throw Lua errors too, but they say **field**, e.g. `admin/client?form=client_list` without `params: {device_mac: "default"}` crashes with `attempt to index field 'device_mac'`. Add the param and it works.

A missing `field` is client-fixable (you forgot a parameter). A missing `global` is not. In Lua, "global" means it lives in the VM's global namespace, populated when the dispatcher registers the handler via `entry(...)`. No HTTP request can fill it in. The handler was registered, but registered wrong.

I brute-forced around 1,400 controller/form combinations looking for any other wireless-write path. Everything else returns the clean `"no such callback"`. So this is the only write endpoint and it's broken.

[...]

PR #203, opened 2026-08-18
fix(c6u): surface decrypted body when response is not valid JSON

dennismdejong:

Relates to #158.

When a write/read fails with an HTTP 500 whose body is still AES-encrypted (no JSON envelope), `response.json()` raises and the real router-side error gets hidden behind the generic `An unknown response` message. This made the Deco BE25 `set_wifi()` failure look like a library bug.

This PR tries to AES-decrypt the raw body on that failure path and appends it to the error message, so the underlying Lua dispatcher traceback is visible:

```
TplinkRouter - TPLinkDecoClient - An unknown response - Expecting value: ... Decrypted response - Failed to execute call dispatcher target ... attempt to index global mode
```

- Only affects the error path (non-JSON body) — no behavior change for working responses
- Guarded by `getattr(self, "_encryption", None)` so non-encrypted clients are untouched

## Tests

- `test_request_error_includes_decrypted_body` — non-JSON body gets decrypted and included in the raised `ClientError`

Full suite: 477 tests pass. flake8 clean.

### DIFF
--- a/test/test_client_c6u.py
+++ b/test/test_client_c6u.py
@@ -1845,6 +1845,32 @@ class TestTPLinkClient(TestCase):
         self.assertEqual(_Stub()._decrypt_response({'data': ''}), {})
         self.assertEqual(_Stub()._decrypt_response({}), {})
 
+    @patch('tplinkrouterc6u.client.c6u.post')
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@dennismdejong Thank you!

PR #208, opened 2026-08-19
fix(c80): enable legacy TLS renegotiation

dennismdejong:

Addresses home-assistant-tplink-router #259.

Python 3.14 / OpenSSL disables legacy renegotiation by default, so the C80 router handshake fails with `UNSAFE_LEGACY_RENEGOTIATION_DISABLED` even when SSL verification is off.

This PR builds an SSL context with `OP_LEGACY_SERVER_CONNECT` and uses it for the C80 session, still honoring `verify_ssl` for certificate checks (previously `verify=False` just bypassed verification without fixing the renegotiation error).

## Tests

- `test_ssl_context_enables_legacy_renegotiation` — option set, verification on by default
- `test_ssl_context_unverified_when_verify_off` — `verify_ssl=False` disables checks

### DIFF
--- a/test/test_client_c80.py
+++ b/test/test_client_c80.py
@@ -799,5 +799,24 @@ class TestTPLinkClient(TestCase):
         self.assertEqual(ipv6_calls['n'], 1)
 
 
+class TestTplinkC80RouterSslContext(TestCase):
+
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@dennismdejong Thank you!

PR #209, opened 2026-08-19
fix(c6u): retry authorize on empty/partial login data block

dennismdejong:

Addresses home-assistant-tplink-router #357.

The router intermittently returns an empty or partial `data` block during login (e.g. `KeyError: data` on Mercusys MR80X), failing the whole integration setup. `authorize()` now retries the full login sequence up to three times before raising.

- Only on the encrypted login path; a genuinely wrong password still fails immediately with the same `Login failed` error.
- Bounded (3 attempts), so a persistent failure is not hidden.

## Tests

- `test_authorize_retries_on_empty_data_block` — succeeds on the retry
- `test_authorize_raises_after_retries_on_empty_data_block` — still raises after 3 attempts

### DIFF
--- a/test/test_client_c6u.py
+++ b/test/test_client_c6u.py
@@ -1866,6 +1866,130 @@ class TestAuthMinimal(TestCase):
             self.client._prepare_data('data')
             self.assertFalse(mock_sign.call_args[0][1], "Should pass False when logged")
 
+    def test_authorize_retries_on_empty_data_block(self) -> None:
+        """The router intermittently returns an empty/partial data block on login
+        (e.g. KeyError 'data' on Mercusys MR80X). authorize() must retry the full
+        login sequence instead of failing on the first attempt."""
+        class Client(TplinkRouter):
+            def __init__(self):
+                super().__init__('http://192.168.0.1', 'password')
+                self.login_calls = 0
+
+            def _request_pwd(self) -> None:
+                self._pwdNN = '00'
+                self._pwdEE = '010001'
+
+            def _request_seq(self) -> None:
+                self._seq = '100'
+
+            def _try_login(self):
+                self.login_calls += 1
+
+                class Resp:
+                    headers = {'set-cookie': 'sysauth=test_sysauth_cookie; path=/'}
+                    status_code = 200
+
+                    def json(self):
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@dennismdejong in https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/357 error shows that there is no data block in response. It looks like that model has a different API for authorization. So adding a retry for a wrong request is not a solution. Plus, if you pass a wrong password, you also get retries with it. It may block your router account

dennismdejong:

Thanks — you are right on both points. I have reworked the PR:

1. **No retry on a wrong request / wrong password.** The retry now triggers **only** on an empty response body (`{}`). Any content-ful failure — including a wrong password, which returns an `errorcode` in the data block — raises immediately without retrying, so the router account lockout cannot be hit. Added `test_authorize_does_not_retry_content_full_failure` to lock this in.

2. **On the "different authorization API" point:** I checked #357. The reporter explicitly says the router "responds ... with an empty body (`{}`) instead of the expected payload" and that it correlates with WAN reconnect/renegotiation, and they explicitly ask for "a retry-on-empty-response mechanism to help mask this kind of transient router-side hiccup." So this is not retrying a malformed request — it is tolerating a transient empty reply from an otherwise-working login. If you believe the MR80X actually needs a different auth flow entirely, I can drop this PR and open an issue to request the API trace instead.

Would you prefer I keep this (empty-response-only retry) or drop it pending an MR80X API trace?

AlexandrErohin:
It needs a deep investigation and testing with a real router, because adding retries may spam the router with "BUSY" responses and may lead to the router freezing. Also, authorization retries may lead to the same error that will occur later in get_status or another method. However, retrying that method may require re-authorization

dennismdejong:

Closing this PR. Agreed with your assessment — a retry in authorize() needs deep investigation and testing against a real MR80X: it could spam the router with BUSY responses, risk freezing, and the same transient failure would likely recur in later methods anyway.

The underlying issue (#357) stays open. The productive next step is an API trace from the reporter (raw login response when it returns the empty body) to determine whether the MR80X needs a different authorization flow rather than a retry.

PR #210, opened 2026-08-19
fix(deco): retry transient read failures (Bad Gateway)

dennismdejong:

Addresses home-assistant-tplink-router #374, #335, #366.

Deco X20/X50/X55 intermittently return a 502 Bad Gateway or non-JSON body on reads (`admin/client?form=client_list`, etc.), which fails the whole integration setup. `_retry_request` now also retries transient `ClientError` reads up to three times with a short delay.

- Writes are never retried, to avoid double-applying a state change.
- Existing `ConnectTimeout` retry behaviour is unchanged.

## Tests

- `test_retry_request_retries_client_error_on_read` — read retries and succeeds
- `test_retry_request_does_not_retry_write` — write fails immediately, single attempt

### DIFF
--- a/test/test_client_deco.py
+++ b/test/test_client_deco.py
@@ -1,5 +1,5 @@
 from unittest import main, TestCase
-from json import loads
+from json import loads, dumps
 from macaddress import EUI48
 from ipaddress import IPv4Address
 from tplinkrouterc6u import (
@@ -11,6 +11,7 @@ from tplinkrouterc6u import (
     IPv4Status,
     LTEStatus,
 )
+from tplinkrouterc6u.common.exception import ClientError
 
 
 class TestTPLinkDecoClient(TestCase):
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
1) retry guard for write-operations never matches real Deco request bodies, so state-changing calls can now be retried and applied multiple times. In TPLinkDecoClient, writes are sent as JSON strings via dumps(...), for example set_wifi() sends {"operation": "write",  ...} and reboot() sends {"operation": "reboot", ...}. But the new retry filter only checks for URL-encoded substrings like operation=write|save|update, so real write payloads fall through the “safe to retry” path and will be retried on any ClientError. That violates the PR’s own safety comment and can duplicate side effects on transient 5xx responses. See the actual write payload construction in tplinkrouterc6u/client/deco.py lines 31-53 and the broken detection logic in tplinkrouterc6u/client/deco.py lines 218-229
2) the new negative test validates the wrong payload format, so it misses the regression above. test_retry_request_does_not_retry_write() passes 'operation=write' directly into _retry_request(...), but production TPLinkDecoClient.request() always forwards JSON produced by dumps(...). As written, the test only proves the string-matcher works for a synthetic format the client never uses. See test/test_client_deco.py lines 819-830 
3) operation: "reboot" also would be rebooted - but it may leed to router break

dennismdejong:

You are entirely right on all three points. I have fixed the PR:

1. **Write detection now matches real Deco payloads.** Deco writes are JSON via `dumps()` (`{"operation": "write"|"reboot"|"logout", ...}`), not URL-encoded. The old `operation=write` substring match never matched them, so writes could be retried and double-applied. The new `_is_read_request()` parses the JSON and only retries `operation` values `read`/`load`.

2. **Non-JSON args (e.g. authorize) are never retried** — treated as non-retryable to stay conservative.

3. **The negative test now uses the real payload format** and covers the reboot case specifically: `test_retry_request_does_not_retry_write` uses `dumps({"operation": "write", ...})`, `test_retry_request_does_not_retry_reboot` uses `dumps({"operation": "reboot", ...})`, and `test_retry_request_does_not_retry_non_json` covers the authorize (no-data) path. Each asserts a single attempt.

Full suite passes (472 tests), flake8 clean.

AlexandrErohin:
This Pull Request does not fix https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/374 - there is a problem with decode.

https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/335 - the router may not supported admin/client?form=client_list

https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/366 here looks like problem with ssl certs, or with router deny 443
So retrying will not help with these problems.

AlexandrErohin:
Also adding retrying needs to test code with some real routers. Because it may break compabilities

dennismdejong:

Closing this PR. After analysis of the three referenced HA issues, retrying does not address their root causes:

- **#374** (Deco X20): deterministic UTF-8 decode error from garbage bytes in a device nickname — not transient; needs a decode fix in aes_decrypt(), not a retry.
- **#335** (Deco X50): mixed — some Bad Gateway (retry could help) but also ReadTimeout (not caught) and possibly an unsupported admin/client?form=client_list endpoint.
- **#366** (Deco X55): SSL/certificate/ReadTimeout on port 443 during authorize — not a read-retry issue.

Thanks for the review — the retry approach does not fit these cases.

Issue #122, opened 2025-12-14
Backup possible?


SjurWarEagle:
Hi,

I'm looking for a way to backup my router settings (TP Link XV800v) and as I use a homeassistant plugin based on this project, I thought you might have an idea if this could support accessting System->Backup->Backup?

<img width="1023" height="922" alt="Image" src="https://github.com/user-attachments/assets/ccab73f0-7a25-4e8b-b8c7-bd7c8faed2cd" />
Thanks a lot!

albertogeniola:
Hi!

I'm pretty sure this answer comes "late", but I wanted to report my experiments. I needed to run a backup action on a Archer VR1200v device. By playing with the UI, I've noticed that the backup can be triggered using the following path:

`GET http://ADMIN_URL/cgi/{hw_prefix}{hw_version}{fw_build}{fw_release}.bin?`

Where:
- hw_prefix: is the device hardware model (e.g. "Archer VR1200v")
- hw_version: is the device hardware version (e.g. "V1")
- fw_build: is the firmware build (e.g. 200714 )
- fw_release: is the firmware build release (2239n)

You can retrieve this info from the Administration page of the admin portal:

<img width="794" height="390" alt="Image" src="https://github.com/user-attachments/assets/87d746e3-d20f-4e80-8651-dfb0cbcb409e" />

In the example above, my "backup path" is the following:
`'/cgi/ArcherVR1200vV12007142239n.bin?'`

So I was able to script the backup creation using the library as follows:
```python
    def run(self) -> None:
        # Login with library
        router = TplinkRouterProvider.get_client("MY_PASSWORD", verify_ssl=False)
        router.authorize()  

        # The backup download path depends on the firmware version name. So we first need to calculate it here.
        fw_info = router.get_firmware()
        hw_parts = re.match("^(Archer VR1200[a-zA-Z0-9]+) (v[0-9]+)", fw_info.hardware_version)
        hw_prefix = hw_parts.group(1).replace(" ","")
        hw_version = hw_parts.group(2).upper()
        fw_parts = re.match(".* Build ([0-9]+) Rel.([0-9a-zA-Z]+)", fw_info.firmware_version)
        fw_build = fw_parts.group(1)
        fw_release = fw_parts.group(2).lower()
        
        backup_path = f"/cgi/{hw_prefix}{hw_version}{fw_build}{fw_release}.bin?"

[...]

PR #211, opened 2026-08-19
feat(vr1200v): add backup_config() to download router settings

dennismdejong:

Addresses #122. Per your guidance ("Other models may have different API for this feature - so it is better to add only for models with same API"), this adds `backup_config()` only to the VR1200v client.

The backup path is built from the firmware version strings exactly as @albertogeniola demonstrated and validated on a real Archer VR1200v:

```
/cgi/{hw_prefix}{hw_version}{fw_build}{fw_release}.bin?
```

- Returns the raw backup file bytes
- Uses the VR1200v auth headers (`TokenID`, `Referer`)
- Raises a clear `ClientException` when the firmware strings do not match the expected format, or on HTTP error
- Handles the hardware version prefix case-insensitively (`v1` / `V1`)

## Tests

- `test_backup_config_builds_path_and_returns_content` — asserts the exact path + auth headers + returned bytes
- `test_backup_config_raises_on_unmatched_firmware` — unmatched version strings
- `test_backup_config_raises_on_http_error` — non-200 response

Full suite passes (471 tests), flake8 clean.

### DIFF
--- a/test/test_client_vr1200v.py
+++ b/test/test_client_vr1200v.py
@@ -1,0 +1,56 @@
+from unittest import TestCase, main
+from unittest.mock import Mock
+
+from tplinkrouterc6u.client.vr1200v import TplinkVR1200vRouter
+from tplinkrouterc6u.common.dataclass import Firmware
+from tplinkrouterc6u.common.exception import ClientException
+
+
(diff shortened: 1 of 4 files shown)


AlexandrErohin:
@dennismdejong Thank you!

PR #212, opened 2026-08-19
fix(encryption): ignore invalid UTF-8 in decrypted payloads

dennismdejong:

Addresses home-assistant-tplink-router #374.

The Deco app can write garbage bytes (invalid UTF-8) into a device nickname, e.g. `"nickname":"m\xab\\^^\x99\xe9\xed"`. When the encrypted `admin/device?form=device_list` response is decrypted and decoded, the strict UTF-8 decode raises `UnicodeDecodeError`, which crashes the whole integration setup.

This PR makes `EncryptionWrapper.aes_decrypt()` decode with `errors="ignore"`, so invalid bytes are dropped and the surrounding valid JSON still parses. This matches the approach the reporter (dmilotic) suggested and tested: `result.decode(errors="ignore")`.

## Tests

- `test_aes_decrypt_ignores_invalid_utf8_bytes` — a payload containing invalid UTF-8 bytes decrypts without raising, preserving the surrounding valid text.

Full suite passes (469 tests), flake8 clean.

### DIFF
--- a/test/test_common_encryption.py
+++ b/test/test_common_encryption.py
@@ -17,6 +17,27 @@ class TestEncryptionWrapper(TestCase):
         ciphertext = wrapper.aes_encrypt(plaintext)
         self.assertEqual(wrapper.aes_decrypt(ciphertext), plaintext)
 
+    def test_aes_decrypt_ignores_invalid_utf8_bytes(self) -> None:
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@dennismdejong Thank you!

PR #214, opened 2026-08-23
Fix MR200 LTE status returning raw strings instead of ints

potat0man:
### Problem

`TPLinkMR200Client.get_lte_status()` assigns values straight from
`req_act()` without the `int()` casts the base `TPLinkMRClient` applies.
Since the CGI response is parsed as plain `key=value` text, every field
arrives as a string.

This breaks three consumers:

- `LTEStatus.network_type_info` and `sim_status_info` look up int-keyed
  tables, so `.get('3')` misses and returns "Unknown network type" /
  "Unknown SIM status"
- Downstream `sig_level * 25` becomes string repetition:
  `'4' * 25` -> `'4444444444444444444444444'`

Reproduced with an MR200 via the Home Assistant integration; the MR600
on the same install is fine because it uses the base client's casts.

Before / after with `networkType='3'`, `simStatus='5'`, `signalStrength='4'`:

| | before | after |
|---|---|---|
| network_type_info | `Unknown network type` | `4G LTE` |
| sim_status_info | `Unknown SIM status` | `SIM unlocked. Authentication succeeded.` |
| sig_level * 25 | `'4444444444444444444444444'` | `100` |

### Change

Adds the missing casts, matching `mr.py`. `total_statistics` uses
`int(float(...))` as the base client does. Adds `test_get_lte_status`
covering the MR200 path — it fails on `main` and passes with the fix.

### DIFF
--- a/test/test_client_mr_200.py
+++ b/test/test_client_mr_200.py
@@ -38,6 +38,39 @@ class TestTPLinkMR200Client(TestCase):
 
         self.assertEqual(result, True)
 
+    def test_get_lte_status(self):
+        values = {
+            '0': {
+                'enable': '1',
+                'connectStatus': '5',
+                'networkType': '3',
+                'simStatus': '5',
+                'signalStrength': '4',
+            },
+            '1': {
+                'totalStatistics': '81234567',
+                'curRxSpeed': '45120',
+                'curTxSpeed': '9310',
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@potat0man Thank you!

PR #215, opened 2026-08-23
Add info message for chosen client.

david9600:
Just an info log message to indicate chosen client.


AlexandrErohin:
@david9600 Thank you!

Issue #140, opened 2026-02-11
TP-Link AX1500 (Access Point mode) – WiFi 2.4GHz and 5GHz entities revert to ON state after being turned off


alespaghetti:
Description

I am experiencing an issue with the TP-Link integration in Home Assistant when using a TP-Link AX1500 configured in Access Point mode.

The integration connects successfully to the device via http and correctly discovers all available entities. However, there is a problem specifically with the WiFi radio entities (2.4 GHz and 5 GHz).

When I turn off either the 2.4 GHz or 5 GHz WiFi entity from Home Assistant, the entity briefly switches to OFF, but after a short time it automatically reverts back to ON, returning to its previous state.

**Expected behavior**

When turning off the 2.4 GHz or 5 GHz WiFi entity from Home Assistant, the WiFi radio should remain OFF until manually turned back ON.

**Actual behavior**

The entity is successfully recognized.

The state changes to OFF when toggled.

After a short time, it automatically switches back to ON.

The WiFi network itself appears to remain active.

**Environment**

Home Assistant version: 2026.2.1

TP-Link integration version:

Device model: TP-Link AX1500

Device mode: Access Point

Firmware version: 1.2.1 Build 20250729 Rel. 73566(4555)

Additional information

The issue seems to occur only when the AX1500 is configured in Access Point mode.
Please let me know if debug logs are needed and I will provide them.

alespaghetti:
> [@alespaghetti](https://github.com/alespaghetti) What is your router hardware version?

Archer AX18 v1.0

david9600:
Hi, the switch behaviour (which I've seen before with another issue linked below) is simply that it shows the newly selected state until the next poll of the router, which returns the actual state. In other words, the 'set' command to the router (sent immediately the switch is changed) is failing.  
I'm not sure which client is used by the AX1500 series, but it may be possible to capture the request and response with network sniffer software, which can then be compared with a working request using browser developer tools when logged into the admin interface of the router.

The earlier similar issue (now fixed for AX55) was [this one](https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/324)

Issue #51, opened 2024-11-03
Archer C24 Support Data


Ill-tableflip-U:
[ArcherC24.zip](https://github.com/user-attachments/files/17610917/ArcherC24.zip)

Let me know if I did anything wrong. I might submit the data for another two unsupported TP-link router models as well.

This is what I did:
click on client list
wireless settings
disable 2.4 GHz. enable 2.4 GHz
disable 5 Ghz. enable 5 GHz.
enable guest 2.4 GHz. disable guest 2.4 GHz
enable guest 5 Ghz. disable guest 5 GHz.
Open settings menu
Reboot
Log in
Logout

Hardware version is Archer C24 V2.0
Firmware version is 1.12.20

AlexandrErohin:
Would be great if you could add support for your router by creating PR with new class that extends AbstractRouter https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client.py#L16

AlexandrErohin:
Hi. The support for the router with same API was added!
Could you please test the client for your router?
```
from tplinkrouterc6u import TplinkRouterProvider
from logging import Logger

router = TplinkRouterProvider.get_client('http://192.168.0.1', 'password')
router.authorize()
router.get_firmware()
router.get_status()
```

Ill-tableflip-U:
output is: `Exception has occurred: KeyError
'33|1,9,0'
  File "C:\Users\Dylan\Downloads\tplinkc24\main.py", line 7, in <module>
    router.get_status()
KeyError: '33|1,9,0'`

And I was unsure about the usage of AbstractRouter, sorry. Would be great if you could provide further details? I have some more models of router that are not yet supported and would love to help get them supported.

AlexandrErohin:
@Ill-tableflip-U I have fixed this error.
Could you download the latest version of tplinkrouterc6u and run again
```
from tplinkrouterc6u import TplinkRouterProvider
from logging import Logger

router = TplinkRouterProvider.get_client('http://192.168.0.1', 'password')
router.authorize()
router.get_firmware()
router.get_status()
```
Please let me know the code result

Ill-tableflip-U:
Alright, thanks for this update. It does seem to be completely functional now (at least with all of the features that this C24 router actually supports. I can toggle the different networks, see the IP address, hardware model etc. The TP-link Archer C24 is now supported.

[...]

PR #218, opened 2026-09-02
feat(mesh): expose the EasyMesh node list through get_mesh_nodes()

ADNPolymerase:
## What

Adds a `MeshNode` dataclass and a `get_mesh_nodes()` method returning the
EasyMesh nodes that the main router reports, so callers can see the mesh itself
and not only the clients attached to it.

Naming follows the agreement in #219, which adds the same concept for the Deco
family: one dataclass and one method across families, with the default on
`AbstractRouter` as the shared contract and each client implementing it for its
own hardware.

Requested in AlexandrErohin/home-assistant-tplink-router#212, where the point was
made that the router does have that table but the integration does not retrieve
it. Related: #51 (rebooting satellites) and #140 (AX55 in mesh mode).

## Why this is a small change

The library already sends this exact request. `TplinkBaseRouter.get_status()`
calls `admin/easymesh_network?form=get_mesh_device_list_all` on every poll, then
uses the answer only to fill `Device.ap_name` and `Device.signal` on connected
clients. The node list itself is parsed and thrown away.

So there is nothing to reverse engineer here, no new endpoint and no new
encryption path: `get_mesh_nodes()` reuses `self._url_easymesh_device_list` and
the client's own `request()` helper, which is what already handles the encrypted
transport on each model.

## Behaviour

- `AbstractRouter.get_mesh_nodes()` raises `NotImplementedError`, mirroring
  `get_ipv6_status`, so clients that do not speak this form stay explicit.
- `TplinkBaseRouter` implements it and honours the existing `_easymesh` flag. A
  router with no mesh answers the form with an error, which is already the signal
  `get_status()` uses to stop enriching clients, so the method returns an empty
  list and does not ask again.
- `role`, `status` and `device_type` are kept as plain `str` rather than promoted
  to enums, because the values differ between firmware generations and an unknown
  value should not break parsing. Happy to change this if you prefer enums.

## Payload differences found between firmware generations

Tested against an Archer BE550, firmware `1.2.4 Build 20260402`, with a five node
mesh (one main router, four satellites, one of them two hops away). Its payload
carries two fields that the AX55 sample already present in `test_client_c6u.py`
does not:

- `signal_level`: the uplink quality as a 1 to 3 bar level. The payload key is
  `signal_strength`, but the Deco family reports dBm under that same name, so the
  bar level gets its own field and `signal_strength` stays reserved for dBm.
  This is the connection quality indicator the router UI renders. Worth noting
  that the same key means dBm inside `mesh_sclient_detail`, so the two should not
  be conflated.
- `support_reboot`: whether the main router can reboot that node. That is
  directly relevant to #51, since the router itself tells you which satellites
  can be rebooted.

The main router entry carries no `parent_mac`, `connect_type`, `mesh_type`,
`signal_level` or `support_reboot`, so those stay `None`. `location` is also
absent on some nodes. `parent_mac` can point at another satellite rather than at
the main router, which is how a multi hop mesh is represented.

## Scope left out on purpose

`get_mesh_nodes()` issues the single list request. Per node data such as
`firmware_version`, `hardware_version`, the mesh `level` and the per band dBm
values live in `admin/easymesh_network?form=mesh_sclient_detail`, which costs one
extra request per node. I left that out to keep the change reviewable, and can
add it behind a flag if you want it.

One consequence worth flagging: `get_status()` already makes the list request, so
a caller that also calls `get_mesh_nodes()` makes it twice. Sharing the result
would mean caching it on the client, which I did not want to introduce silently.
Tell me if you would rather have that.

## Tests

`python -m unittest discover ./test` passes, 532 tests, and flake8 is clean.

Three tests were added covering the parse, the unsupported router path and an
empty response, plus one on `TPLinkDecoClient` for the `NotImplementedError`
default. Because `TestTPLinkClient` is subclassed by the AX72, V1_11 and SG test
cases, these run against `TplinkRouterSG` as well, which is the client BE series
routers are matched to.

The fixture is an anonymised capture: every MAC, IP address and user chosen name
was replaced, and the topology and field names were preserved.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,14 @@ try:
     except NotImplementedError:
         pass
 
+    # Get EasyMesh nodes (empty list when the router runs no mesh)
+    try:
+        for node in router.get_mesh_devices():
+            print(f"{node.name:24} {node.model:16} {node.macaddr} {node.ipaddr:16s} "
+                  f"{node.role:16} clients={node.client_num}")
+    except NotImplementedError:
+        pass
+
     # Optional methods (not on every client)
     try:
         reservations = router.get_ipv4_reservations()
@@ -127,6 +135,7 @@ Not every method is available on every client. Methods below `get_ipv6_status` t
 | get_status |   | Gets status about the router info including wifi statuses and connected devices info | [Status](#status) |
 | get_ipv4_status |   | Gets WAN and LAN IPv4 status info, gateway, DNS, netmask | [IPv4Status](#IPv4Status) |
 | get_ipv6_status |   | Gets WAN IPv6 status info, gateway, DNS, site prefix (c6u/SG, MR/EX/VR, C80-style; others raise `NotImplementedError`) | [IPv6Status](#IPv6Status) |
+| get_mesh_devices |   | Gets the EasyMesh nodes reported by the main router. Returns an empty list on routers without EasyMesh; clients that do not implement it raise `NotImplementedError` | [[MeshDevice]](#MeshDevice) |
 | get_ipv4_reservations |   | Gets IPv4 reserved addresses (static) | [[IPv4Reservation]](#IPv4Reservation) |
 | add_ipv4_reservation | macaddr: str, ipaddr: str, comment: str = '', enable: bool = True | Adds an IPv4 DHCP address reservation (Archer / LuCI clients) |   |
 | delete_ipv4_reservation | macaddr: str | Deletes an IPv4 DHCP address reservation (Archer / LuCI clients) |   |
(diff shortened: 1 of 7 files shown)


ADNPolymerase:
Verified end to end against real hardware, not only the fixture.

Archer BE550, firmware 1.2.4 Build 20260402, five node EasyMesh network, client resolved by the provider as `TplinkRouterSG`. MAC addresses below are redacted, the topology is untouched.

```
ROLE         MODEL            MAC                UPLINKS THROUGH    CLIENTS SIGNAL REBOOT
-----------------------------------------------------------------------------------------
main         Archer BE550     AA-BB-CC-00-00-01  -                       34   None   None
satellite    RE330            AA-BB-CC-00-00-02  AA-BB-CC-00-00-03        1      3   True
satellite    Archer AX55      AA-BB-CC-00-00-03  AA-BB-CC-00-00-01        9      2   True
satellite    RE500X           AA-BB-CC-00-00-04  AA-BB-CC-00-00-01        9      3   True
satellite    RE330            AA-BB-CC-00-00-05  AA-BB-CC-00-00-01        2      3   True
```

Three things this confirms beyond the unit tests:

- The multi hop case is real, not hypothetical. The first RE330 uplinks through the AX55, which itself uplinks through the main router, so `parent_mac` genuinely describes a tree rather than a flat list of satellites.
- `signal_strength` and `support_reboot` are absent from the main router entry and present on every satellite, which is why they are typed as optional and left at `None` rather than defaulted.
- `client_num` is live and per node. Between two runs a few minutes apart the per node figures moved while the network total stayed in the same range, which is what roaming clients look like, so callers should treat it as a sample rather than a stable attribute.

david9600:
To test the proposed changes, I copied them to my dev environment (for both this library and the HA integration), then tested with a 3-node setup (AX55 main node, AX23 wired satellite, MR70X wireless satellite).
It seems to be well thought out, and on checking values reported in the device tracker entities, appears to be working as expected.

I agree with the choice of 'signal_strength' for dBm measurements, and signal_level for bar-type indication (my AX55 main node seems to report a scale of 1-5 for a wireless node).  The existing per-client dBm values would be the signal strength received at the router from each device (rather than reported by client itself).

On adding the integration entry, I did receive the following warning (obviously not critical, but will need fixing at some stage).
```
WARNING (MainThread) [homeassistant.helpers.frame] Detected that custom integration 'tplink_router_dev' calls `device_registry.async_get_or_create` with a deprecated `via_device` parameter; 
use `via_device_id` instead at custom_components/tplink_router_dev/device_tracker.py, line 139: async_add_entities(new_tracked). 
This will stop working in Home Assistant 2027.8.0, please report it to the author of the 'tplink_router_dev' custom integration
```

@AlexandrErohin  The approach I would suggest, in relation to the overlap of easymesh_device_list query with my original device-to-node implementation, is that this PR (and the one in the integration) be merged first, and I will then create a PR to relocate the device-to-node processing from get_status to get_mesh_nodes.

ADNPolymerase:
Thanks for testing this on a 3-node setup. Your merge order works for me: once #218 and #402 are in, relocating the device-to-node processing into get_mesh_nodes is yours.

On the via_device warning: it does come from the mesh node trackers. It is more than a one-line swap, though:
- `via_device_id` only exists from HA 2026.8.0, and hacs.json still declares 2024.4.0, so replacing the key outright would break older installs.
- `via_device_id` needs the parent device already registered, otherwise HA refuses the entity, so the order nodes are added in starts to matter.
- The client-as-device option already on main (#397) passes `via_device` the same way, so it will raise the same warning once enabled.

It keeps working until 2027.8.0, so I would rather fix both call sites in one follow-up PR than patch only the mesh one here. @AlexandrErohin, would you prefer a version check, or raising the minimum HA version to 2026.8.0?

On signal_level: the docs describe it as a 1 to 3 bar level, which only reflects what I saw on a BE550. Your AX55 goes up to 5, so the range depends on the model. Only the wording is affected, no code relies on the range. Should I update both PRs now, or would you rather handle the wording yourselves, for instance in the follow-up?

AlexandrErohin:
@david9600 I have merged this Pull Request. Would be great to have your Pull Request before we moved to home assistant repo. Thank you!

AlexandrErohin:
@ADNPolymerase Thank you for your Pull Request! Abour warning - I think raising the minimum HA version to 2026.8.0 is not good as that release was only month ago. https://analytics.home-assistant.io/ A lot of users have older versions, I think we should wait to raise the minimum HA version

david9600:
I'm working on relocating the client-to-node details - may take a day or two, as it's not quite as straightforward as get_status.  Will also update the comments about the signal bar level so they're generic.

david9600:
PR #235 submitted.

PR #219, opened 2026-09-03
feat(deco): expose mesh units via get_mesh_nodes()

edrianolima:
## Problem

`TPLinkDecoClient.get_firmware()` reads `admin/device?form=device_list`, which
returns **every unit of the mesh**, but keeps only the master's firmware:

```python
for item in self.devices:
    if item.get('role') != 'master' and len(self.devices) != 1:
        continue          # <- everything about the other units is dropped
```

The data is already fetched and stored in `self.devices` (`reboot()` uses it to
build its `mac_list`), and then discarded. Consumers such as the Home Assistant
integration therefore see a Deco mesh as a single device, and a slave unit that
drops off or degrades is invisible.

## Change

Adds a `MeshNode` dataclass and `TPLinkDecoClient.get_mesh_nodes()`. It reuses
the cached `self.devices`, so **no extra request is made** — it only stops
throwing the data away.

Per unit: `nickname`, `role`, `model`, IP, MAC, hardware/firmware version,
`internet_status`, `group_status`, and for slaves the backhaul metrics
(2.4/5 GHz signal in dBm, rx/tx rates), the parent MAC and the wired port count.

Follows the existing `Device` dataclass conventions: `EUI48`/`IPv4Address` in
private fields with `macaddr`/`ipaddr` string properties alongside
`macaddress`/`ipaddress` typed ones.

A master reports no `signal_strength`, `rx_rate_list`, `tx_rate_list` or
`previous` — it has no uplink of its own — so those fields are `None` there.
The firmware mixes ints and numeric strings across these fields, hence the
`_to_int` guard.

## Testing

Verified against a 3-unit **Deco X50** mesh, firmware
`1.8.0 Build 25102213 Rel. 43970`:

```
[MASTER] Living Room   192.168.68.1
[slave ] Kids          192.168.71.250   5G -50 dBm   rx 960   tx 1441 Mbps
[slave ] Main Bedroom  192.168.71.249   5G -57 dBm   rx 1152  tx 1152 Mbps
```

Both slaves report the master as parent; the master reports none.

- `test_get_mesh_nodes` — real captured payload covering both master and slave,
  asserting every field and that the master's backhaul fields are `None`
- `test_get_mesh_nodes_reuses_cached_device_list` — asserts no second request is
  issued after `get_firmware()`

Full suite passes (512 tests), flake8 clean.

## Note

I have a companion PR ready for `home-assistant-tplink-router` that turns this
into one device per mesh unit with backhaul sensors. Happy to adjust the shape
of `MeshNode` first if you would prefer it modelled differently.

### DIFF
--- a/test/test_client_deco.py
+++ b/test/test_client_deco.py
@@ -6,6 +6,7 @@ from tplinkrouterc6u import (
     TPLinkDecoClient,
     Connection,
     Firmware,
+    MeshNode,
     Status,
     Device,
     IPv4Status,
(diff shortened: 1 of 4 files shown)


edrianolima:
Noticed #218 landed a day before this one, solving the same problem from the
other side — worth flagging so you do not have to work out the overlap yourself.

They are complementary rather than competing:

|  | #218 | #219 (this one) |
|---|---|---|
| Family | EasyMesh (Archer) | Deco |
| Client | `TplinkBaseRouter` | `TPLinkDecoClient` |
| Endpoint | `admin/easymesh_network?form=get_mesh_device_list_all` | `admin/device?form=device_list` |

A Deco does not answer the EasyMesh form, and the Archer models have no
`device_list`, so neither PR covers the other's hardware. Amusingly we both
found the same thing independently: the request was already being made and the
node list parsed and discarded.

The overlap is in naming, not behaviour — `MeshDevice` / `get_mesh_devices()`
against `MeshNode` / `get_mesh_nodes()` for what a caller sees as the same
concept. That is yours to arbitrate, and I am happy to rename mine to match
#218, or to align both behind one shared shape if you would rather have a single
mesh API across families. Say which and I will push the change here.

One field-level difference worth knowing if you do decide to converge: on the
Deco payload, `signal_strength` is a **dBm value** per band, while #218 reports
that on EasyMesh the same key is a **1-3 bar level** (with dBm living in
`mesh_sclient_detail` instead). So a shared dataclass would need either separate
fields or an explicit unit, otherwise the same attribute would carry two
different meanings depending on the router.

ADNPolymerase:
Agreed on all of it, and since you offered to rename yours, let me settle it the other way so we stop trading concessions: `MeshNode` / `get_mesh_nodes()` wins, I will rename #218.

On the shared contract, #218 already adds the default on `AbstractRouter`, so one dataclass and one method with each client implementing it for its own family needs nothing beyond that rename.

Your units point is the real obstacle, and it has a clean answer that #402 already uses on the integration side. Do not share the key. `signal_strength` keeps a single meaning across families, dBm, populated by Deco and left `None` on EasyMesh where dBm lives in `mesh_sclient_detail`. The 1-3 bar level gets its own field, `signal_level`, populated by EasyMesh and `None` on Deco. One key, one unit. Same rule for anything that collides later: a new field rather than a shared key with a per family meaning.

That also settles the question of the two families blurring together. They stay separable by client class, by disjoint `role` values (`main_router` / `satellite_router` against `master`), and by the fields only one of them populates.

ADNPolymerase:
Done and pushed. #218 now exposes `MeshNode` / `get_mesh_nodes()`, with the bar level in `signal_level` and `signal_strength` left free for your dBm.

One thing worth flagging, since #403 does the same probe: the `hasattr` / `getattr` check on the integration side has to be renamed along with it. Left on the old name it resolves to `None` forever once the library ships the new one, and the entities are silently never created, with nothing in the logs.

Only the dataclass field is renamed; the router payload key stays `signal_strength`.

edrianolima:
Thanks for the rename.

The probe warning does not apply here: #403 already calls `get_mesh_nodes` in both places, in `coordinator.py` and in `__init__.py`. It was never on the old name.

Worth flagging the one overlap between our two library PRs that does not announce itself. We both add `class MeshNode` to `common/dataclass.py`, your hunk at base line 127, mine at line 15. They do not touch, so git merges both cleanly and the module ends up holding two classes of that name. The later definition wins, which is yours, and my `deco.py` passes `nickname`, `internet_status` and `signal_2g`, none of which exist on it. Two green PRs, no conflict reported on that file, `TypeError` on the first real Deco call.

The other two overlaps do announce themselves and will conflict on whichever merges second: `__init__.py` and `test/test_client_deco.py`. The second is a contradiction rather than just text, since `test_get_mesh_nodes_not_supported` asserts that `TPLinkDecoClient.get_mesh_nodes()` raises `NotImplementedError`, which #219 implements.

None of this is actionable from either side yet. Both PRs are correct against `main` as it stands, and whether this ends as one shared `MeshNode` or two separate classes is @AlexandrErohin's call on his own API. Whoever merges second does the rebase once that is decided.

One correction while we are here: my nodes carry no `signal_strength` field at all. The Deco reports the backhaul per band, so it is `signal_2g` / `signal_5g`, both in dBm. The key you left free is one I do not use. What still collides is `name` against `nickname`, and `is_main_router` against `is_master`.

ADNPolymerase:
You are right on both counts, and I checked rather than take it on faith.

The probe warning does not apply: #403 has `hasattr(client, "get_mesh_nodes")` in `__init__.py` and `hasattr(router, "get_mesh_nodes")` in `coordinator.py`. It was never on the old name.

And your nodes carry no `signal_strength` field, it is `signal_2g` / `signal_5g`. What misled me is that both payloads do use that key, a per band dict on Deco and a scalar bar level on EasyMesh, and neither ends up in a field of that name. So the collision I described was not real. The rename to `signal_level` stands on its own, a 1 to 3 bar level is not a strength, but not for the reason I gave.

Your silent overlap is real and it reproduces. Our two PRs overlap on three files, and only `test/test_client_deco.py` conflicts. `dataclass.py` auto merges into a module holding two `class MeshNode`, yours at line 19 and mine at 187, so mine shadows yours. `__init__.py` auto merges to `MeshNode` imported twice, so that one does not announce itself either. Your own call site then raises `TypeError: MeshNode.__init__() got an unexpected keyword argument 'nickname'`.

Worth stating plainly for @AlexandrErohin: the usual signal, one routine looking conflict in a test file, under reports this by a lot.

Agreed the shared or separate call is yours. If it lands on shared, the two remaining collisions can be settled between us so only that decision is left: `name` rather than `nickname`, the generic word over the Deco specific one, and one predicate covering both role vocabularies instead of two, whichever name you prefer, implemented as `role in ('main_router', 'master')`.

edrianolima:
Agreed on both of your convergence terms, so if @AlexandrErohin picks one shared
`MeshNode` there is nothing left for the two of us to argue about:

- `name` over `nickname`. The generic word is right for a shared dataclass, and
  `nickname` only reads naturally because it is what the Deco payload happens to
  call the field.
- One predicate over two, implemented as `role in ('main_router', 'master')`.
  Name it whichever way you prefer; `is_main_router` is the better of the two we
  each shipped, since "master" is the vendor's word and not a concept the caller
  should have to know.

You were right that I got `__init__.py` wrong, and I should not have asserted it
without checking, so I ran the merge rather than take your word for that either:

```
$ git merge mine/feat/deco-mesh-nodes && git merge his/feat/mesh-devices
CONFLICT (content): Merge conflict in test/test_client_deco.py
Auto-merging tplinkrouterc6u/__init__.py
Auto-merging tplinkrouterc6u/common/dataclass.py
```

One conflict, in a test file. The two files that decide whether the package works
merge silently, and this is what they leave behind:

```
$ grep -n '^class MeshNode' tplinkrouterc6u/common/dataclass.py
19:class MeshNode:      # nickname, is_master
187:class MeshNode:     # name

$ grep -c 'MeshNode,' tplinkrouterc6u/__init__.py
2
```

The definition at 187 shadows the one at 19, so `deco.py` then constructs a class
that has no `nickname`, and the first real Deco call raises `TypeError`.

@AlexandrErohin, the part worth flagging: the only signal here is one
ordinary-looking conflict in a test file. CI on the merge commit would not catch
the rest, because nothing fails until a Deco is actually queried.

Happy to do the rebase on my side whenever the shared-or-separate call is made.

AlexandrErohin:
@edrianolima @ADNPolymerase Thank you for your Pull requests! I think it is better to use one class for MeshNode. So we can use it in HomeAssistant. I think to merge the first one https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/218 and would be great if @edrianolima could update this Pull Request after

edrianolima:
Sounds good. I'll rebase #219 onto #218 once it lands.

Unless you'd rather decide differently, I'll adopt what @ADNPolymerase and I
converged on earlier in this thread: `name` rather than `nickname` on the shared
dataclass, and a single predicate `role in ('main_router', 'master')` covering
both vocabularies.

No rush on my side; ping me when #218 is in.

AlexandrErohin:
@edrianolima I agree

AlexandrErohin:
I have released some features today and I need to back off from merging for some days to see if I need to release fixes for today release.

AlexandrErohin:
@edrianolima I have merged https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/218 Would be great if you could update your Pull Request. Thank you!

edrianolima:
Done, rebased onto `main` with #218 in. Uses the shared `MeshNode` with the agreed `name` and `is_main_router`, plus the Deco-only fields (`signal_2g`/`signal_5g` in dBm, per-band rates, `wired_ports`, hw/fw versions).

Heads-up: git flagged only the `test_client_deco.py` conflict, while `dataclass.py` and `__init__.py` auto-merged into two `class MeshNode` in one module, and nothing fails until a Deco is actually queried. Also dropped `test_get_mesh_nodes_not_supported`, which asserted the opposite of this PR.

620 tests, flake8 clean. #403 waits on this one.

AlexandrErohin:
@edrianolima Thank you!

PR #222, opened 2026-09-06
Add extended WAN support & USB modem support (MR client).

david9600:
This adds support for DHCP release/renew on Ethernet WAN link, and support for USB modem (WAN backup link).
Status updates are added into get_status in MR client as separate requests, and disabled after initial probe if router doesn't support the function.
Ideally, these requests would be made conditional on user-selected options from the Config Flow, but I'm not sure of the best way to pass this information to the client.

A new method set_ewan_connect allows ethernet WAN link to be disconnected (DHCP release) and connected (DHCP renew).  This can be used in an automation, for example when resetting a link which hasn't fully initialised for ipv4 & ipv6 after a router reboot.

I've also renamed the existing wan_aux_acts/values (the 'aux' was originally chosen to allow other functions to be added, but that code block is now just for ipv6).

If the PR is accepted, I will raise a PR on the integration for switch and sensors.

### DIFF
--- a/tplinkrouterc6u/client/mr.py
+++ b/tplinkrouterc6u/client/mr.py
@@ -101,6 +101,8 @@ class TPLinkMRClientBase(AbstractRouter):
         self._ee = None
         self._seq = None
         self._ipv6_support = True
+        self._wan_extd_support = True
+        self._wan_usb_support = True
         self._url_rsa_key = 'cgi/getParm'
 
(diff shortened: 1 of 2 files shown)


david9600:
I've realised it will be better to do the USB modem together with changes for dual WAN in a separate PR, so I will close this one and resubmit just for the E-WAN release/renew.

Issue #59, opened 2024-12-10
TL-WR844N supported ?


barbuslex:
Hi,

TL-WR844N is it supported ? i have not see this reference in documentation.

Thanks

AlexandrErohin:
Hi, There wasnt info about TL-WR844N. If you have it - would be great if you could test

barbuslex:
I will receive 2x TL-WR844N models next week... I will try...
Is there a special configuration to do for unsupported models?

AlexandrErohin:
No special configuration

barbuslex:
Ok I'll test next week and I'll give you feedback

barbuslex:
I have try to add it today, i have this error : 
![image](https://github.com/user-attachments/assets/7401b256-665c-411e-88ea-504ada841bd0)

AlexandrErohin:
So it is not supported. Would be great if you could add support for your router. Here are the examples how to do it https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/CONTRIBUTING.md

barbuslex:
Yes of course i Can try...

What IS the serie of TL-WR844N model ? TL series ?

AlexandrErohin:
@barbuslex You may implement methods for this class https://github.com/AlexandrErohin/TP-Link-Archer-C6U/blob/main/tplinkrouterc6u/client/c6v4.py, It looks like API the same for these routers

AlexandrErohin:
@barbuslex Hi! Could you try the HA integration latest version please?

barbuslex:
I have tried : 

![Image](https://github.com/user-attachments/assets/5a4f6029-3074-4871-a9c9-ad998d547caa)

It seems in TL-WR844N there are not have username : 

![Image](https://github.com/user-attachments/assets/3440ee2c-3348-4646-bd2a-6b7ea1dc805d)

dennismdejong:
> *This was generated by AI during triage.*

## Status update: support has been added

Support for the TL-WR844N has since been added in `main` via PR #179: the router is picked up by the **c80 client** (the status response identifies `modelName TL-WR844N`, and parsing is covered by a test, including the no-wifi status shape).

**On the "no username" point:** this is not a blocker — the c80 client uses password-only encryption (`_encrypt_password`); the default `admin` username is not part of the login exchange.

@barbuslex — could you please test the latest version (5.32.1+) again?

```python
[...]

PR #223, opened 2026-09-06
fix(re330,c80): treat 0000-marker error replies as plain text

dennismdejong:

Addresses issue #59 (TL-WR844N).

## Root cause (confirmed with a real response from the reporter)

The WR844N answers `get_firmware()` with **`00006\r\n`** (raw hex `30303030360d0a`) — the `0000` OK/error marker plus an error code — not an encrypted payload. Both the re330 and c80 clients fed such replies to `b64decode()` inside `_decrypt_data()`, crashing with `Invalid base64-encoded string` (5 chars is not a multiple of 4).

The provider currently selects the **re330** client for the WR844N (both classes share the `0000`-style API), so the fix covers both.

## Fix

- Share `is_valid_base64()` in `common/helper.py`.
- `_decrypt_data()` in re330 and c80 now returns any `0000`-prefixed reply that is **not** valid base64 as plain text (covers `00000\r\n...`, bare `00000`, and error codes like `00006\r\n`).
- Encrypted payloads are still decrypted unchanged.

## Tests

- helper: `test_is_valid_base64` (valid / wrong length / CRLF marker / non-base64 chars / padding)
- re330 + c80: `test_plain_text_with_crlf_is_returned`, `test_error_code_marker_is_not_fed_to_base64` (`00006\r\n`), `test_bare_ok_marker_is_not_fed_to_base64`, `test_valid_base64_is_still_decrypted`

Full suite passes (523 tests), flake8 clean.

### DIFF
--- a/test/test_client_c80.py
+++ b/test/test_client_c80.py
@@ -847,5 +847,24 @@ class TestTplinkC80RouterSslContext(TestCase):
         self.assertIs(client._session.verify, sentinel)
 
 
+class TestTplinkC80RouterDecryptData(TestCase):
+
+    def test_plain_text_with_crlf_is_returned(self) -> None:
+        client = TplinkC80Router('http://192.168.0.1', 'password')
+        text = '00000\r\nid 0|1,0,0\r\nmodelName TL-WR844N'
+        self.assertEqual(client._decrypt_data(text), text)
+
+    def test_bare_ok_marker_is_not_fed_to_base64(self) -> None:
+        """TL-WR844N can answer with a bare '00000' (no CRLF, not base64);
+        it must be returned as-is instead of crashing on b64decode (#59)."""
+        client = TplinkC80Router('http://192.168.0.1', 'password')
+        self.assertEqual(client._decrypt_data('00000'), '00000')
+
+    def test_valid_base64_is_still_decrypted(self) -> None:
+        client = TplinkC80Router('http://192.168.0.1', 'password')
+        encrypted = client._encryption.aes.aes_encrypt('hello world')
+        self.assertEqual(client._decrypt_data(encrypted), 'hello world')
+
+
 if __name__ == '__main__':
     main()
--- a/test/test_client_c80.py
+++ b/test/test_client_c80.py
@@ -860,6 +860,13 @@ class TestTplinkC80RouterDecryptData(TestCase):
         client = TplinkC80Router('http://192.168.0.1', 'password')
         self.assertEqual(client._decrypt_data('00000'), '00000')
 
+    def test_error_code_marker_is_not_fed_to_base64(self) -> None:
+        """The real TL-WR844N answer for get_firmware is '00006\\r\\n'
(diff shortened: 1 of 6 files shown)


AlexandrErohin:
@dennismdejong Thank you! I think we need real response from WR844N for the request. This Pull request still decrypt sting if it starts 00000foo

dennismdejong:
@barbuslex could you please provide us a real response so we can validate the PR?

barbuslex:
Hi @dennismdejong, I have replied here in #59.

dennismdejong:

You are right on both points, and thanks for the review.

1. **The fix needs to be in re330, not (only) c80.** The reporter just tested 5.32.1: the provider now selects `TplinkRE330Router` for the TL-WR844N (both classes use the same `00000`-style API and probe, so re330 wins in provider order). re330 `_decrypt_data()` has no plain-text handling at all — it always calls `aes_decrypt()`, so it hits the same `b64decode` crash.

2. **The `00000foo` gap is real.** My `_is_valid_base64()` treats a string that starts with `00000` and continues with base64-valid characters as ciphertext, so it would still be decrypted. A string-level check cannot reliably separate those two cases.

So we agree we need the **real response** from the WR844N. I have asked the reporter (on #59) to log the raw `get_firmware()` response (repr + hex) — the current error shows 5 characters, consistent with a bare `00000` OK marker, but we need confirmation.

Proposed next step once we have it: add a shared plain-text guard for `00000`-prefixed replies to both re330 and c80 `_decrypt_data()`, and use the captured response to decide the exact shape. Happy to rework this PR (or close it and open one that covers re330) based on that.

dennismdejong:

The reporter confirmed the real response: `get_firmware()` on the TL-WR844N returns **`00006\r\n`** (hex `30303030360d0a`) — the `0000` marker plus error code `6`, no encrypted payload. The client was feeding it to `b64decode()` and crashing.

I have reworked this PR per our discussion:
- Shared `is_valid_base64()` in `common/helper.py`
- Both re330 (the client the provider now selects for WR844N) and c80 `_decrypt_data()` return any `0000`-prefixed non-base64 reply as plain text
- Covered `00000\r\n...`, bare `00000`, and the confirmed `00006\r\n` error code

**Open point for you:** even with this fix, `get_firmware()` will not return firmware data on the WR844N — the router answers the `0|1,0,0` request with error code `6`. Is there a different request the re330/c80 client should use for firmware on this model, or is firmware via this API simply unsupported on the WR844N? If the latter, get_firmware would need to handle the error-code reply gracefully.

dennismdejong:

Closing this PR in favour of a dedicated `TplinkWR844NRouter` client. The reporter confirmed on the real WR844N that data requests must be sent **plain** (unencrypted); the encrypted request is rejected with error code 6. A shared fallback in re330/c80 would risk those working clients, so a dedicated client (per the existing `beta` wr844n.py, corrected) is the cleaner approach.

PR #225, opened 2026-09-09
feat(wr844n): add dedicated plain-text data client

dennismdejong:

Addresses #59 (TL-WR844N).

## Root cause (confirmed with a real device)

The TL-WR844N uses the same C80-style block API, but answers data requests (firmware, status) with a **plain-text** block. It rejects an encrypted body with error code `00006\r\n`. The existing re330/c80 clients always encrypt data requests, so `get_firmware()`/`get_status()` fail on the WR844N. The provider currently falls back to `TplinkRE330Router` for it.

Confirmed by the reporter:
- encrypted `0|1,0,0` -> `00006\r\n`
- plain `0|1,0,0` (with token) -> `00000\r\nid 0|1,0,0\r\n... modelName TL-WR844N ... softVer 1.15.20 ...`

## Fix

Add a dedicated `TplinkWR844NRouter(TplinkC80Router)` that sends data requests **plain** with the session token (`asyn=1`), matching the working web-UI call. It reuses the authorize handshake (which is encrypted and works) and adds `get_firmware()` + `get_status()` parsing the plain blocks. Registered in the provider before C80/RE330 so it is selected for the WR844N.

## Tests

- `test_supports`
- `test_get_firmware_handles_plaintext_response` (real `softVer 1.15.20 Build 260611 Rel.30987n(4555)`)
- `test_get_status_uses_dhcp_client_block`

Full suite passes (513 tests), flake8 clean.

### DIFF
--- a/test/test_client_wr844n.py
+++ b/test/test_client_wr844n.py
@@ -1,0 +1,109 @@
+from ipaddress import IPv4Address
+from unittest import TestCase, main
+
+from macaddress import EUI48
+
+from tplinkrouterc6u import Connection
+from tplinkrouterc6u.client.wr844n import TplinkWR844NRouter
+from tplinkrouterc6u.common.dataclass import Device, Firmware, Status
+from tplinkrouterc6u.common.exception import ClientException
+
+
+FIRMWARE_RESPONSE = ('00000\r\nid 0|1,0,0\r\nfullName 300Mbps%20Wi-Fi%20Router\r\nfacturer TP-Link\r\n'
+                     'modelName TL-WR844N\r\nmodelVer 1.0\r\n'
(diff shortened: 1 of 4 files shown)


AlexandrErohin:
@dennismdejong Thank you for your Pull Request and for helping with https://github.com/AlexandrErohin/TP-Link-Archer-C6U/issues/59 . I have rewieved your code and see that it only fixes get_firmware and get_status for TL-WR844N. Other methods for TL-WR844N wont work and same routers with TL-WR844N wont be supported. I have added more general approach based on your research https://github.com/AlexandrErohin/TP-Link-Archer-C6U/commit/3244f2ff4e75e1cfbefddd77b078b6fd123ea525

PR #226, opened 2026-09-10
Add support for IOT, 6G and MLO for EX series routers

amosyuen:
Add support for IOT, 6G and MLO networks for EX series routers.
- Set enable status for each band
- Set band to MLO for devices on MLO
- Devices in an unknown wifi will still be counted towards total and have unknown type

Tested on my HB810 routers in HA


amosyuen:
Will fix https://github.com/AlexandrErohin/home-assistant-tplink-router/issues/362 for Ex series routers

AlexandrErohin:
@amosyuen Thank you!

PR #229, opened 2026-09-11
fix: populate TL-SG108E LAN IPv4 address

prvashisht:
## Problem

`TPLinkSG108EClient.get_status()` populates the switch port totals and LAN MAC, but leaves `lan_ipv4_addr` unset even though the management address is available from the existing IP settings page.

This prevents consumers from using the existing `Status.lan_ipv4_addr` field for the switch.

## Fix

* Populate `Status._lan_ipv4_addr` from the TL-SG108E IP settings.
* Keep the lookup best-effort so an IP settings failure does not break otherwise valid switch status.
* Add regression coverage for the populated and unavailable cases.

No public API, aggregate semantics, port status behavior, or switch configuration is changed.

## Testing

Tested on:

* TL-SG108E v6.0

Physical switch validation:

* management IPv4 is returned on repeated `get_status()` calls;
* LAN MAC remains populated;
* physical port total remains 8;
* connected-port count remains consistent with per-port link state;
* repeated reads do not change switch configuration.

Also verified:

* `pytest test/test_client_sg108e.py` — 13 passed
* `pytest test` — 537 passed
* flake8 clean
* `git diff --check` clean

### DIFF
--- a/test/test_client_sg108e.py
+++ b/test/test_client_sg108e.py
@@ -161,14 +161,16 @@ class TestTPLinkSG108EClient(TestCase):
                 'link_status': [6, 6, 0, 5, 5, 6, 5, 6, 0, 0],
             },
         })
-        # The MAC lookup is best-effort and must not break the port counts.
+        # The MAC/IP lookups are best-effort and must not break the port counts.
         client.device_info = Mock(side_effect=Exception('system info unavailable'))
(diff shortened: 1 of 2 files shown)


AlexandrErohin:
@prvashisht Thank you! Maybe cache IP like self._lan_mac ? So we remove extra request for IP every get_status update

prvashisht:
Implemented the suggested LAN-IP cache follow-up in #232.

PR #231, opened 2026-09-12
Fix get_sms() when message content contains equals sign

quectobit:
I ran into this issue when using this library for something because the line could be something like `content=Visit https://example.com/page?id=1`, but I'm a bit of a python noob so let me know if there's better syntax for this.

This library is really awesome btw, thanks for making it.


AlexandrErohin:
@quectobit Thank you!

PR #232, opened 2026-09-16
fix: cache TL-SG108E LAN IPv4 address

prvashisht:
## Problem

Follow-up to #229.

`TPLinkSG108EClient.get_status()` now exposes the switch management IPv4, but resolving it on every status update adds an IP-settings request to each polling cycle.

## Fix

Cache the management IPv4 per client instance using the same first-resolution pattern as the existing LAN MAC cache.

`get_ipv4_status()` remains a fresh IP-settings read and does not use the cached status value.

The lookup remains best-effort, and no public API, port behavior, aggregate semantics, or switch configuration is changed.

## Testing

Tested on:

* TL-SG108E v6.0

Physical validation confirmed:

* repeated `get_status()` calls return management IPv4 `192.168.68.103`;
* only the first `get_status()` performs the IP-settings request;
* later `get_status()` calls reuse the cached address;
* explicit `get_ipv4_status()` performs a fresh IP-settings request;
* LAN MAC remains populated;
* physical-port and connected-port counts remain unchanged.

Also verified:

* focused SG108E tests pass
* full pytest suite passes
* flake8 clean
* `git diff --check` clean


AlexandrErohin:
@prvashisht Thank you!

PR #235, opened 2026-09-21
Relocate mesh client data collection.

david9600:
This removes mesh client data collection from get_status and adds it to get_mesh_nodes where it can share the router's node list response.

Data collected:
* ap_name for all mesh clients.
* signal_strength for clients on a satellite node.

The data is returned as a list to the coordinator, where it is used to enrich the device data from get_status.
Tested on a 3-node EasyMesh network (including coordinator changes which I can submit as a PR once #402 has been merged on the integration).


AlexandrErohin:
@david9600 Thank you for the Pull Requests! I think it would be great to have full info for mesh devices. So what about a separate method like this one?
```
get_device_mesh_info(list[Device]) -> list[Device]
```
The existing get_mesh_nodes would be only for MeshNode.
And the new one would be for mesh devices? But to reduce requests, the new method would be called after get_status – and from get_status a list[Device] would be passed to get_device_mesh_info to enrich this info in Device.

david9600:
There are definitely pros and cons.
It really depends how many people want just node information, how many people are more interested in which nodes the clients are connected to, and how many people want both.

The main goal of adding the client-to-node queries to the get_mesh_nodes method was to avoid retrieving the easymesh_device_list twice (to obtain node mac addresses which are needed for the nclient queries).  A get_device_mesh_info may not be convenient as a library method on its own, as it would either have to do the initial query to get the node mac addresses, or they would have to be passed to it.

Admittedly, adding the collection of client information to get_mesh_nodes means the client queries (one per node) are sent to the router, even if the user is interested only in the mesh backbone and not the client associations.

If a significant number of users want only one of the two functions, the ideal would be to allow selection at the Functional Customisation panel.

AlexandrErohin:
@david9600 Unfortunately you added second return value for get_mesh_nodes and it greak compability with this merged Pull Request https://github.com/AlexandrErohin/TP-Link-Archer-C6U/pull/219/changes . Maybe add to MeshNode parameter devices - so get_mesh_nodes would stay same
```
def get_mesh_nodes(self) -> list[MeshNode]
```
What do you think?

david9600:
OK, that one wasn't merged at the time I submitted the PR.
It might be easier if I close this one, and we just leave the mesh client requests where they are now, in get_status.

AlexandrErohin:
@david9600 Agree. It looks like it is a good way

